Issue #11, opened 2019-06-23
ATMega328P Support


jonahbron:
Opening another issue here for stuff specifically related to this repo

Related to: https://github.com/Rahix/avr-hal/issues/2

@Rahix In your first comment on that issue you said the first place to start is working on the chip patch.  I see that there is already an atmega328p patch file, but it has fewer _`include`s.  My guess is that it's incomplete in its current form.

> Luckily, the ATmega328P is very similar to the already implemented ATmega32U4. This should mean that you can for the most part copy the patches from that chip.

That sounds good... the only issue is that I don't know where to find the differences between the 32u4 and the 328p.  Is that info I can find in the data sheet?  Or documented elsewhere on the Microchip website?  Again please pardon my extreme ignorance here.

jonahbron:
Reading through the [atmega328p.adtf](https://github.com/Rahix/avr-device/blob/master/vendor/atmega328p.atdf), the [patch YAML files](https://github.com/Rahix/avr-device/tree/master/patch), and the [ATMega328 datasheet](http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-7810-Automotive-Microcontrollers-ATmega328P_Datasheet.pdf), I'm beginning to see relationships between their contents.  For example, I see that the YAML files seem to target XML elements in their adtf files by `[name]` attribute.  And the info I see in the adft files seems to have corresponding details in the datasheet.

There are a few things that are still mysterious, principally some of the underscore patch actions in the patch file.  The following are ones I see in some of the YAML files:

- `_write_constraint` - found [here](https://github.com/Rahix/avr-device/blob/master/patch/common/timer/tc0.yaml#L5)
- `_replace_enum` - found [here](https://github.com/Rahix/avr-device/blob/master/patch/common/timer/tc0.yaml#L11)

Neither of those are mentioned in the [stm32-rs README](https://github.com/stm32-rs/stm32-rs/#device-and-peripheral-yaml-format).  Where can I find documentation on their behavior?  The `_replace_enum` command is particularly confusing because I see no enums _at all_ at all the adtf's.

At a higher level, it seems to me like the objective here is to add patches such that when applied to the atdf file, has an output that matches what's described in the datasheet.  Is that roughly accurate?

Rahix:
> My guess is that it's incomplete in its current form.

Yes

> Is that info I can find in the data sheet? Or documented elsewhere on the Microchip website?

[...]

PR #12, opened 2019-06-23
Added missing patches for atmega328p

jonahbron:
Added missing patches for atmega328p by copying the commons from atmega32u4, omitting PLL because the 328p lacks that feature


jonahbron:
resolves #11

Rahix:
Thank you!

PR #13, opened 2019-06-23
common: Fix a few patches

jonahbron:


patch/common/wdt.yaml
```diff
@@ -30,3 +30,5 @@ WDT:
       CYCLES_64K:    [5, "- 64K (65536) cycles, ~0.5s"]
       CYCLES_128K:   [6, "- 128K (131072) cycles, ~1s"]
       CYCLES_256K:   [7, "- 256K (262144) cycles, ~2s"]
+      CYCLES_512K:   [8, "- 512K (524288) cycles, ~4s"]
+      CYCLES_1024K:  [9, "- 1024K (1048576) cycles, ~8s"]
```
Rahix: Actually, those values are not in here for a reason ;)  If you look closely, the `WDPL` field is only 3 bit wide, meaning `7`(`0b111`) is the biggest value to fit in there.  The additional values in the datasheet use the `WDPH` bit which is unfortunatly separate from the rest.  This is why I instead listed `512K` and `1024K` in `CYCLES_2K` and `CYCLES_4K` respectively.

But as you doing this change shows, the way I did it was probably too confusing.  If you want to, I suggest renaming `CYCLES_2K` to `CYCLES_2K_512K` and `CYCLES_4K` to `CYCLES_4K_1024K`?

patch/atmega328p.yaml
```diff
@@ -1,3 +1,6 @@
 _include:
-  - "common/wdt.yaml"
   - "common/ac.yaml"
+  - "common/timer/tc0.yaml"
+  - "common/twi.yaml"
+  - "common/usart.yaml"
+  - "common/wdt.yaml"
```
Rahix: Those changes were already merged in bfbd5bc006ad745037c004d90405ac216cb64509.  Maybe you need to rebase your branch ontop of my current master?

jonahbron: Yeah I think Github is glitching out here, I wasn't sure why it was showing up.  Will try rebasing.

patch/common/wdt.yaml
jonahbron: Ahhh that makes sense, I did not read the descriptions thoroughly enough.  I'll remove the new lines and make the suggested changes.

patch/common/wdt.yaml
```diff
@@ -22,11 +22,11 @@ WDT:
         description: "Watchdog Timer Prescaler - High Bit"
         bitRange: "[5:5]"
     WDPL:
-      CYCLES_2K:     [0, "- 2048 cycles, ~16ms/512K (524288) cycles, ~4s if WDPH is set"]
-      CYCLES_4K:     [1, "- 4096 cycles, ~32ms/1024K (1048576) cycles, ~8s if WDPH is set"]
-      CYCLES_8K:     [2, "- 8192 cycles, ~64ms"]
```
jonahbron: Indented remaining lines to keep vertical alignment

jonahbron:
@Rahix All comments addressed, thanks for your feedback.

Issue #15, opened 2019-06-26
Startup


Rahix:
Right now, the startup is not as pretty as I would like it to be.  Ideally, we should add a `#[entry]` procedural macro as [cortex-m](https://github.com/rust-embedded/cortex-m-rt/blob/master/macros/src/lib.rs#L85) did.

Rahix:
We should also think about pulling in the startup code written by @shepmaster in their [rust-arduino-blink-led-no-core-with-cargo](https://github.com/shepmaster/rust-arduino-blink-led-no-core-with-cargo) project instead of relying on avr-gcc.  Though this does not have a high priority for now.

jonahbron:
@Rahix What would be the material benefits?

Rahix:
Right now, not that much.  But looking into the future, it would mean independence from `avr-gcc` and `avr-libc` and ideally the ability to build the entire binary using just llvm tools.  Though I think we are still a long way away from that.

Rahix:
Oh, and now that I put some work into #19, I can tell that using custom startup code would also make the interrupt macro a lot less painful to implement.

Rahix:
I think it makes sense to add a stub `#[entry]` macro now, before publishing on crates.io, so we can later just switch the underlying implementation without breaking user code.

PR #19, opened 2019-06-30
Implement interrupts

Rahix:
This PR adds a new feature-flag `rt` which, when enabled, adds the `#[interrupt]` procedural macro to define an interrupt handler.  Unlike the implementation in cortex-m, this version needs an attribute which is the name of the chip the interrupt is for.  In code, an interrupt handler might look like this:
```rust
#![feature(abi_avr_interrupt)]

#[avr_device::interrupt(atmega32u4)]
fn INT6() {
    // Do Something
}
```


Rahix:
@peacememories, you originally suggested this over in my old `atmega32u4` crate.  What do you think of the implementation I added in this PR?

peacememories:
Hmm, is there a specific reason you’re using macro arguments instead of generating one macro for each chip? Since most people would use the chip-specific module anyway, this would make the api more like the more used cortex one, and reduce the boilerplate.
Of course we don’t need to be hell-bent on following the cortex-m api, but unless I’m missing something this seems like it would be pretty simple?

peacememories:
Hmm, i guess multiple interrupt macros are a bit weird too ^^

Rahix:
Yes, I thought about this as well.  In an ideal world, I'd love to have something similar to

```rust
#[avr_device::atmega32u4::interrupt]
fn USART1_RX() {
}
```

This is unfortunately not possible because macros always have to live in the crate root.  It would also be quite awkward to implement because you'd have to generate the proc-macro defintions (Like code-generating code-generators? This has some serious meta-spirits ...).  In the end I chose the way proposed here as it is the easiest to actually implement for now.

Although, the situation will improve in the future!  Once we replace the startup code from `avr-gcc` with our own, this all can be simplified to look exactly like in cortex-m (ref #15):

```rust
#[avr_device::interrupt]
fn USART1_RX() {
}
```

peacememories:
I wonder... do you think we can encode semantics like “when this interrupt fires the uart has incoming data” in the interrupts? Would that make sense? Why I would be interested in doing this: in c I would just start reading data from uart on an interrupt, but in rust I’d use `read` on the uart which (afaict) also Checks if there is actually data in the register. This is good for safety, but technically when the uart interrupt fires I have a guarantee that there is data available. Same for adc samples etc.
This is not something super pressing, but I don’t know where else bouncing ideas like this around would fit better ^^

Along the same lines: I’m trying to implement a (static) runtime for futures for avr, and I have some concerns about the zero-cost-ness of the abstraction as well. But there I’ll first have to check if rust/llvm is able the optimize out dynamic dispatch in certain situations ^^
Nothing to do with this pr though ^^’

Rahix:
I think this deserves its own issue over in `avr-hal`.  Would you mind creating one?

Back to this PR though, any objections to merging the current proposal with the intention of getting it closer to cortex-m once we have our own startup?

peacememories:
No objections here :)

PR #21, opened 2019-07-05
Enabled write access to SPI2X for compatible chips

jonahbron:


Rahix:
Thanks!  I am currently busy with other things so I don't have much time for avr-rust.  But please don't take my absence as lacking interest!

PR #22, opened 2019-08-14
Use sh as an interpreter for gen-intr-lut.sh

dflemstr:
No `/bin/bash` features are in use and it is not guaranteed that bash is always installed in `/bin/bash` (i.e. might be `/usr/bin/bash`)

### DIFF
--- a/gen-intr-lut.sh
+++ b/gen-intr-lut.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 # Generate a lookup table function for interrupts of all supported chips
 set -e
 


Rahix:
I've recently learned from a NixOS guy that we should really use `/usr/bin/env bash` for maximum compatibility.

BTW, `"$@"` is bash-only, so we *are* using bash-specifics ;)

Rahix:
I pushed commit 6f0b1cc91482 ("gen-intr-lut: Switch to /usr/bin/env bash as interpreter").  Does this work for you?

dflemstr:
Seems like it *is* working for me, thanks!

PR #27, opened 2020-07-24
Use llvm_asm!() instead of asm!()

lperlaki:


src/lib.rs
```diff
@@ -16,7 +16,7 @@
 //! * `atmega64`
 //! * `attiny85`
 #![no_std]
-#![feature(asm)]
+#![feature(asm, llvm_asm)]
```
Rahix, line 19: As in `avr-hal`, we don't need the `asm` feature anymore for now as
everything is using the old llvm-style asm syntax.

Rahix:
Thanks again!

PR #29, opened 2020-07-25
Add patches for ADC peripherals

lperlaki:
Patch needed for ADC implementation
currently only `atmega328p`

[avr-hal/11](https://github.com/Rahix/avr-hal/issues/11)


Rahix:
From a quick look, this patch should also be applied for `ATmega1280`, `ATmega32U4`, and `ATmega64`.  Would you mind adding that to this PR?

PR #32, opened 2020-07-31
Add support for ATmega2560

couchand:
Basically just copied from the ATmega1280 files, which seems reasonable enough as they're more or less the same chips.  Simple testing done through `avr-hal`.


Rahix:
Looks good, thanks!

PR #33, opened 2020-08-04
Add support for ATtiny88

couchand:
This one was a fair bit more work, due to the many minor discrepancies between this chip and others already supported.  I'm not familiar enough with the patching environment to know the best way to abstract the common bits, but I'd be happy to work to clean things up if given some pointers.

Thanks!

### DIFF
--- a/patch/attiny88.yaml
+++ b/patch/attiny88.yaml
@@ -1,0 +1,127 @@
+_include:
+  - "common/ac.yaml"
+  - "common/spi.yaml"
+  - "common/wdt.yaml"
+
+  - "timer/attiny88.yaml"
+
+ADC:
+  ADMUX:
+    _modify:
+      MUX:
+        description: "Analog Channel Selection Bits"
+    MUX:
+      _replace_enum:
+        ADC0: [0, "ADC Single Ended Input pin 0"]
+        ADC1: [1, "ADC Single Ended Input pin 1"]
+        ADC2: [2, "ADC Single Ended Input pin 2"]
+        ADC3: [3, "ADC Single Ended Input pin 3"]
+        ADC4: [4, "ADC Single Ended Input pin 4"]
+        ADC5: [5, "ADC Single Ended Input pin 5"]
+        ADC6: [6, "ADC Single Ended Input pin 6"]
+        ADC7: [7, "ADC Single Ended Input pin 7"]
+        TEMPSENS: [8, "Temperature sensor"]
+        ADC_VBG: [14, "Internal Reference (VBG)"]
+        ADC_GND: [15, "0V (GND)"]
+    REFS0:
+      _replace_enum:
+        INTERNAL: [0, "Internal 1.1V Voltage Reference"]
+        AVCC: [1, "AVcc Reference"]
+  ADCSRA:
+    ADPS:
+      _replace_enum:
+        PRESCALER_2: [1, "Prescaler Value 2"]
+        PRESCALER_4: [2, "Prescaler Value 4"]
+        PRESCALER_8: [3, "Prescaler Value 8"]
+        PRESCALER_16: [4, "Prescaler Value 16"]
+        PRESCALER_32: [5, "Prescaler Value 32"]
+        PRESCALER_64: [6, "Prescaler Value 64"]
+        PRESCALER_128: [7, "Prescaler Value 128"]
+  ADCSRB:
+    ADTS:
+      _replace_enum:
+        FREE: [0, "Free Running mode"]
+        AC: [1, "Analog Comparator"]
+        INT0: [2, "External Interrupt Request 0"]
+        TC0_CMA: [3, "Timer/Counter0 Compare Match A"]
+        TC0_OVF: [4, "Timer/Counter0 Overflow"]
+        TC1_CMB: [5, "Timer/Counter1 Compare Match B"]
+        TC1_OVF: [6, "Timer/Counter1 Overflow"]
+        TC1_CE: [7, "Timer/Counter1 Capture Event"]
+CPU:
+  _modify:
+    CLKPR:
+      access: read-write
+    OSCCAL:
+      access: read-write
+    PORTCR:
+      access: read-write
+    PRR:
+      access: read-write
+  CLKPR:
+    CLKPS:
+      _replace_enum:
+        PRESCALER_1: [0, "Prescaler Value 1"]
+        PRESCALER_2: [1, "Prescaler Value 2"]
+        PRESCALER_4: [2, "Prescaler Value 4"]
+        PRESCALER_8: [3, "Prescaler Value 8"]
+        PRESCALER_16: [4, "Prescaler Value 16"]
+        PRESCALER_32: [5, "Prescaler Value 32"]
+        PRESCALER_64: [6, "Prescaler Value 64"]
+        PRESCALER_128: [7, "Prescaler Value 128"]
+        PRESCALER_256: [8, "Prescaler Value 256"]
+  SMCR:
+    SM:
+      _replace_enum:
+        IDLE: [0, "Idle"]
+        ADC: [1, "ADC Noise Reduction"]
+        PDOWN: [2, "Power-down"]
+EEPROM:
+  EECR:
+    EEPM:
+      _replace_enum:
+        ATOMIC: [0, "Atomic (erase and write in one operation)"]
+        ERASE: [1, "Erase only"]
+        WRITE: [2, "Write only"]
+EXINT:
+  EICRA:
+    ISC0:
+      _replace_enum:
+        LOW: [0, "The low level of INT0 generates an interrupt request"]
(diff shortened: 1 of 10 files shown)


Rahix:
If it helps, the patching is essentially what is documented in [`svdtools`' README](https://github.com/stm32-rs/svdtools#device-and-peripheral-yaml-format) + a few local additions found in [our README](https://github.com/Rahix/avr-device#adding-a-new-chip).  I'm [working on integrating our changes upstream](https://github.com/stm32-rs/svdtools/pull/31), though so we can hopefully switch away from our patched version at some point.

Trying to pull out common bits is a bit difficult, because of the messy state the vendor files are in.  I usually create a backup of the old patched SVD files (`svd/*.patched.svd`), then start trying to rewrite the patches to extract common bits and check along the way what is changed and whether all of that is as intended.

To be perfectly honest, though: I'd be okay with taking this in pretty much as-is and doing the refactoring later if at all.  It's not like the patches are a thing that will be changed much in the future and when adding a new chip you have to check whether each of them is correct anyway, so the gain from putting in all that work is quite minimal ...

patch/attiny88.yaml
```diff
@@ -0,0 +1,127 @@
...
+  TWSR:
+    _modify:
+      TWS:
+        access: read-only
+    TWPS:
+      _replace_enum:
+        PRESCALER_1: [0, "Prescaler Value 1"]
+        PRESCALER_4: [1, "Prescaler Value 4"]
+        PRESCALER_16: [2, "Prescaler Value 16"]
+        PRESCALER_64: [3, "Prescaler Value 64"]
```
Rahix, line 127: Was there a reason you couldn't use the common TWI patch?  Maybe we can change the patch in some way to make it compatible?

patch/attiny88.yaml
```diff
@@ -0,0 +1,127 @@
...
+        LOW: [0, "The low level of INT0 generates an interrupt request"]
+        TOGGLE: [1, "Any logical change on INT0 generates an interrupt request"]
+        FALLING: [2, "The falling edge of INT0 generates an interrupt request"]
+        RISING: [3, "The rising edge of INT0 generates an interrupt request"]
+    ISC1:
+      _replace_enum:
+        LOW: [0, "The low level of INT1 generates an interrupt request"]
+        TOGGLE: [1, "Any logical change on INT1 generates an interrupt request"]
+        FALLING: [2, "The falling edge of INT1 generates an interrupt request"]
+        RISING: [3, "The rising edge of INT1 generates an interrupt request"]
```
Rahix, line 99: I'd bunch the two fields together:

```yaml
ISC?:
  _replace_enum:
      LOW: [0, "The low level of INTx generates an interrupt request"]
      TOGGLE: [1, "Any logical change on INTx generates an interrupt request"]
      FALLING: [2, "The falling edge of INTx generates an interrupt request"]
      RISING: [3, "The rising edge of INTx generates an interrupt request"]
```

Ideally this might also be reusable for the other chips at some point ...

patch/attiny88.yaml
couchand: The common patch chokes on 
```yaml
TWI:
  _modify:
    TWCR:
      access: read-write
```
with
```
Traceback (most recent call last):
  File "patch/svdpatch.py", line 834, in <module>
    main()
  File "patch/svdpatch.py", line 827, in main
    process_device(svd, root)
  File "patch/svdpatch.py", line 806, in process_device
    update_fields)
  File "patch/svdpatch.py", line 732, in process_peripheral
    process_peripheral_modify(ptag, rspec, rmod)
  File "patch/svdpatch.py", line 218, in process_peripheral_modify
    rtag.find(key).text = value
AttributeError: 'NoneType' object has no attribute 'text'
Makefile:26: recipe for target 'svd/attiny88.patched.svd' failed
make: *** [svd/attiny88.patched.svd] Error 1
```

It would seem that this particular model's ATDF doesn't specify read-only for the control register.

patch/attiny88.yaml
couchand: > I'd bunch the two fields together

That's how I wrote it at first, but it didn't end up replacing the enum values for ISC1.

... *goes off to look through `svdpatch.py`* ...

I *think* the change to make is this, but I'm really out of my element here:

```patch
diff --git a/patch/svdpatch.py b/patch/svdpatch.py
index c70c3a3..1281e9c 100644
--- a/patch/svdpatch.py
+++ b/patch/svdpatch.py
@@ -629,7 +629,7 @@ def process_field_enum(pname, rtag, fspec, field, usage="read-write"):
     derived = None
     for ftag in iter_fields(rtag, fspec):
         name = ftag.find('name').text
-        if derived is None:
+        if derived is None or replace_if_exists:
             enum = make_enumerated_values(name, field, usage=usage)
             enum_name = enum.find('name').text
             enum_usage = enum.find('usage').text
```

patch/attiny88.yaml
couchand: Ok, I got this one fixed by looking back at your changes in e0d031a3 and adapting them.

patch/attiny88.yaml
couchand: I've applied the above in fddd2d8, let me know if that looks right to you.

patch/attiny88.yaml
Rahix: Ah, right, I was thinking about this while preparing the svdpatch changes for upstream.  Looks like it came back to bite earlier than I thought!

patch/attiny88.yaml
Rahix: Hmm, I don't think that's quite right.  You're essentially short-circuiting the derived enum mechanism.  Please try this patch instead:

```diff
From 150847c8079e9f4cbd8542051c5e37e9472d5a75 Mon Sep 17 00:00:00 2001
From: Rahix <rahix@rahix.de>
Date: Wed, 5 Aug 2020 06:05:21 +0200
Subject: [PATCH] svdpatch: Fix _replace_enum when used on derived fields

Make sure the enum is replaced properly in the case of derived
enumerated values.

Signed-off-by: Rahix <rahix@rahix.de>
---
 patch/svdpatch.py | 30 +++++++++++++++++-------------
 1 file changed, 17 insertions(+), 13 deletions(-)

diff --git a/patch/svdpatch.py b/patch/svdpatch.py
index 0d1f592be450..7d75c1c9d541 100644
--- a/patch/svdpatch.py
+++ b/patch/svdpatch.py
@@ -632,24 +632,28 @@ def process_field_enum(pname, rtag, fspec, field, usage="read-write"):
         field = field["_replace_enum"]
         replace_if_exists = True
 
-    derived = None
+    derived, enum, enum_name, enum_usage = None, None, None, None
     for ftag in iter_fields(rtag, fspec):
         name = ftag.find('name').text
-        if derived is None:
+
+        if enum is None:
             enum = make_enumerated_values(name, field, usage=usage)
             enum_name = enum.find('name').text
             enum_usage = enum.find('usage').text
-            for ev in ftag.iter('enumeratedValues'):
-                ev_usage_tag = ev.find('usage')
-                ev_usage = ev_usage_tag.text if ev_usage_tag is not None else 'read-write'
-                if ev_usage == enum_usage or ev_usage == "read-write":
-                    if replace_if_exists:
-                        ftag.remove(ev)
-                    else:
-                        print(pname, fspec, field)
-                        raise SvdPatchError(
-                            "{}: field {} already has enumeratedValues for {}. Use '_replace_enum' to overwrite."
-                            .format(pname, name, ev_usage))
+
+        for ev in ftag.iter('enumeratedValues'):
+            ev_usage_tag = ev.find('usage')
+            ev_usage = ev_usage_tag.text if ev_usage_tag is not None else 'read-write'
+            if ev_usage == enum_usage or ev_usage == "read-write":
+                if replace_if_exists:
+                    ftag.remove(ev)
+                else:
+                    print(pname, fspec, field)
+                    raise SvdPatchError(
+                        "{}: field {} already has enumeratedValues for {}. Use '_replace_enum' to overwrite."
+                        .format(pname, name, ev_usage))
+
+        if derived is None:
             ftag.append(enum)
             derived = make_derived_enumerated_values(enum_name)
         else:
-- 
2.27.0
```

Rahix:
Would it be okay for you if I send 8432a0a5a6f980ea416fbeb668f4a8bedc21b95f upstream to svdtools (with proper attribution of course)?

couchand:
> Would it be okay for you if I send 8432a0a upstream to svdtools (with proper attribution of course)?

Certainly.

> Hmm, I don't think that's quite right. You're essentially short-circuiting the derived enum mechanism.

I was a bit confused about how the derived enums work.  Is that what ends up creating the type alias if two different fields share the same set of variants?

Your patch seems reasonable, I'll apply it and check that everything looks right.  Thanks!

couchand:
Ok, great.  I've confirmed that your patch doesn't change any of the existing chip's patched SVDs, and it generates a shared struct definition for `ISC?` here which looks consistent with similar constructs in other places.

Please take another look, thanks.

PR #36, opened 2020-08-06
Makefile improvements

couchand:
Adds a check target to run `cargo check --all-features`, which I've found useful to verify I'm not totally breaking things.

Also adds `all` and `clean` to the `.PHONY` list.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ RUSTUP_TOOLCHAIN ?= nightly
 PATCHES := $(foreach chip, $(CHIPS), $(wildcard patch/$(chip).yaml))
 DEPS := $(foreach patch, $(PATCHES), $(patsubst patch/%.yaml, .deps/%.d, $(patch)))
 
-.PHONY: chips deps $(CHIPS) vector
+.PHONY: chips deps $(CHIPS) vector all clean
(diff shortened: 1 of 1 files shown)


couchand:
Added `rm src/generic.rs` to the clean target.  Probably not necessary generally since it will get clobbered, but as I'm hacking around on `svd2rust` I want to be absolutely sure it's getting the new version, and it seems like the right thing to do anyway.

Makefile
```diff
@@ -56,11 +56,17 @@ macros/src/vector.rs: src/devices/*/interrupt.rs
 	@echo -e "\tGEN-VECTOR\t>macros/src/vector.rs"
 	@./gen-intr-lut.sh $^ >$@
 
+check: all
+	@echo -e "\tCHECK"
+	@cargo check --all-features
+
```
Rahix, line 62: I'd rather keep the makefile just for generating the files, not for actually invoking any kind of compilation, linting, etc.

couchand: That's a fair point.

I only added it after forgetting to pass `--all-features` and hitting the compile warning one too many times.

couchand: It's gone.

Rahix:
Thank you very much!

PR #45, opened 2020-08-17
Make the PINx registers writeable

couchand:
> Writing a logic one to PINxn toggles the value of PORTxn, independent on the value of DDRxn. Note that the SBI instruction can be used to toggle one single bit in a port.


Rahix:
Makes sense, thanks!

PR #47, opened 2020-08-17
Add support for the ATtiny84

couchand:
Another of my favorite tinies!  This is what I was working on when I noticed the discrepancies in the ATtiny85 implementation.

This thing has so many ADC channels, it was really painful to enter them all.  I think they're all right, but I wish there was a good way to test.  I guess I need to build an ADC channel test bench.

Ran into another svdpatch bug on this one - if a register doesn't have any fields defined you can't add any.


patch/attiny84.yaml
```diff
@@ -0,0 +1,221 @@
...
+        bitOffset: 0
+        bitWidth: 6
+        access: read-write
+      REFS:
+        description: "Reference Selection Bits"
+        bitOffset: 6
+        bitWidth: 2
+        access: read-write
+    MUX:
+      _write_constraint: enum
```
Rahix, line 58: Does this work?  I think `_write_constraint` can only be used in an `_modify` block?

patch/attiny84.yaml
```diff
@@ -0,0 +1,221 @@
...
+        ADC6_ADC5_20X: [61, "Differential Inputs Positive ADC6 Negative ADC5 20x Gain"]
+        ADC6_ADC7_1X: [30, "Differential Inputs Positive ADC6 Negative ADC7 1x Gain"]
+        ADC6_ADC7_20X: [31, "Differential Inputs Positive ADC6 Negative ADC7 20x Gain"]
+
+        ADC7_ADC3_1X: [56, "Differential Inputs Positive ADC7 Negative ADC3 1x Gain"]
+        ADC7_ADC3_20X: [57, "Differential Inputs Positive ADC7 Negative ADC3 20x Gain"]
+        ADC7_ADC6_1X: [62, "Differential Inputs Positive ADC7 Negative ADC6 1x Gain"]
+        ADC7_ADC6_20X: [63, "Differential Inputs Positive ADC7 Negative ADC6 20x Gain"]
+        ADC7_ADC7_1X: [38, "Differential Inputs Positive ADC7 Negative ADC7 1x Gain"]
+        ADC7_ADC7_20X: [39, "Differential Inputs Positive ADC7 Negative ADC7 20x Gain"]
```
Rahix, line 132: I'm not sure if it is a good idea to intentionally encode enum values here that can't even fit into the field.  This will be non-obvious to downstream code, where one could miss needing to manually set the upper bits elsewhere.  Instead, I would suggest encoding the values in an enum over in the HAL, [like I did for `ATmega32U4`](https://github.com/Rahix/avr-hal/blob/master/chips/atmega32u4-hal/src/adc.rs#L7-L29) for example.

patch/attiny84.yaml
```diff
@@ -0,0 +1,221 @@
...
+        TWO_WIRE_MASTER: [3, "Two-wire mode (Master). Uses SDA and SCL pins."]
+    USICS:
+      NO_CLOCK: [0, "No Clock/Software clock strobe"]
+      TC0:      [1, "Timer/Counter0 Compare Match"]
+      EXT_POS:  [2, "External, positive edge"]
+      EXT_NEG:  [3, "External, negative edge"]
+  USISR:
+    _modify:
+      USIDC:
+        access: read-only
```
Rahix, line 221: Can we make the USI patch common for all the `ATtiny` chips or are the differences too big for that to be feasible?

patch/svdpatch.py
```diff
@@ -376,10 +376,13 @@ def process_peripheral_add_int(ptag, iname, iadd):
...
-        if ftag.find('name').text == fname:
-            raise SvdPatchError('register {} already has a field {}'
-                                .format(rtag.find('name').text, fname))
+    if parent is None:
+        parent = ET.SubElement(rtag, 'fields')
+    else:
+        for ftag in parent.iter('field'):
+            if ftag.find('name').text == fname:
+                raise SvdPatchError('register {} already has a field {}'
+                                    .format(rtag.find('name').text, fname))
```
Rahix, line 385: Makes sense, should I add that to my [patchset](https://github.com/stm32-rs/svdtools/pull/31) for upstreaming as well?

patch/attiny84.yaml
couchand: On the tiny84 they did the right thing and moved `ADLAR` out of `ADMUX` so that the whole `MUX` field fits in one register.

patch/attiny84.yaml
couchand: It's identical, I can make that change.

patch/svdpatch.py
couchand: Works for me.

patch/attiny84.yaml
couchand: I don't know, I'll go double check.

couchand:
Thanks for the feedback.  I'll get back to this over the next few days and push an updated version.


> Heya, sorry for the delayed response this time ...

No problem at all!

> You're doing this quite thoroughly, as always, which is really nice! Thanks a lot!

Thank you!  My dream is one day to replace the datasheets with docs.rs.

> One thing I do have to carefully ask, though, as you're sending quite a few of these PRs lately: Are those all microcontrollers you're actively using or are you just doing this for fun?

I love the tinies!  I use the x4, x5, and x8 for all sorts of random projects.  The mega2560 is my "behemoth controller for when I'm sitting at my desk and don't want to worry about pin allocation", and otherwise I mostly traffic in the tinies.  (As an aside, I also love the ATtiny10 but that is a totally different animal...)

> The reason I'm asking: My original intention was to keep avr-device constrained to just the MCUs which are actually in use by someone as to not have the crate grow bigger and bigger without anyone downstream benefitting. The lifetime of avr-device in its current form is limited, as it will eventually become too big (if people keep wanting to use it, that is) where we'll have to reorganize into smaller crates with each one supporting a single family. I'm just trying to keep that day as far away as possible without hurting anyone ;)

Totally reasonable.  One thing I've discovered by having this all in once place that I think I otherwise would have missed is how similar-yet-different each chip is, and where the lines of change are.  I'm happy to help refactor the code to keep pushing back that split-everything-up deadline.

patch/svdpatch.py
couchand: As an aside, good job sorting through the new bug there.

patch/attiny84.yaml
couchand: You're right, it's not relevant here.  I can skip the `_replace_enum` as well, since the field is new.

patch/attiny84.yaml
couchand: Done, and I pulled out a couple more, too.

couchand:
I've updated to latest, please take another look, thanks!

Rahix (approved):
Nice work, this looks great!

PR #54, opened 2020-10-04
Add atmega328pb

jkristell:
This adds support for the atmega328pb device. It's basically a atmega328p with 4 more pins available as PORTE and additional TWI, SPI, USART and timers.

I added new functionality to svdtool for stripping prefix from registers and fields from the additional peripherals to make them work with avr-hal without too many modifications. My WIP changes to the hal are in this branch: https://github.com/jkristell/avr-hal/tree/add-atmega328pb

I need confirmation if this is the right way to do it, and then I need to look at the timers as well.


Rahix:
Hey, thanks for the PR!  One note upfront:  Please don't do any changes to the `svdpatch` script in this repo anymore, [I'm in the process of removing that](https://github.com/Rahix/avr-device/pull/51) in favor of the upstream [`svdtools`](https://github.com/stm32-rs/svdtools) version.  That means, if any changes are necessary, please send a PR over there :)

patch/atmega328pb.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - "0"
+TWI1:
+  _strip_end:
+    - "1"
+  TWCR:
+    _strip_end:
+      - "1"
+  TWSR:
+    _strip_end:
+      - "1"
```
Rahix, line 28: Hmm, I'm unsure about this ... This isn't the first time we're fighting with the odd naming scheme of AVR registers but I'm inclined to try sticking with the original names as much as possible.  In my opinion, it's better to have the macros from `avr-hal` flexible enough to take whatever names Atmel gave them than to try patching them all to fit.

The reason is simple: This way we have to deal with it once in the macro instead of everytime a new ATDF file with even more creative postfixes shows up.  Even worse, sometimes the numbers aren't at the end, but embedded somewhere in the middle of a register name.  Additionally, this makes cross-referencing C code with Rust easier as the names stay the same everywhere.

---

I know the HAL currently isn't very consistent about this; `twi` makes the HAL crate spell out all register names entirely while `spi` doesn't expose the names at all currently.  I've recently updated at least the `usart` to a flexible macro that uses [`paste`](https://crates.io/crates/paste) (see [`avr-hal-generic/src/serial.rs`](https://github.com/Rahix/avr-hal/blob/master/avr-hal-generic/src/serial.rs)).  This seems to be working well so at some point `twi` and `spi` should probably follow suit.

jkristell:
> Hey, thanks for the PR! One note upfront: Please don't do any changes to the `svdpatch` script in this repo anymore, [I'm in the process of removing that](https://github.com/Rahix/avr-device/pull/51) in favor of the upstream [`svdtools`](https://github.com/stm32-rs/svdtools) version. That means, if any changes are necessary, please send a PR over there :)

The upstream version has support for _strip_end, so that should be fine. I can rebase this PR on your upstream-svdtool and test.

patch/atmega328pb.yaml
jkristell: I'm not sure either. The pros for doing it this way is that you don't have to write "specific but identical" code for the peripheral instances. In this special case the registers are suffixes with index, and the fields are too for the second device, but not the first. So you have to supply register field names as well :-/ and as you said, if the index is in the middle of the name you can't supply a empty suffix to the macro.

The only other devices I have worked with are stm32s, and they are doing the fixups in the "device" crate, so that's the primary reason I did it this way here. But I do think it's nice to fix the naming in some cases as they tend to be inconsistent for no really good reason and it just make reading and using them harder.

jkristell:
changing base in the PR wasn't such a good idea :) Could you open the PR again so that I can change the base back to master?

Rahix:
Hmm, I can't because the branch is gone :( I'll try recreating it ...

jkristell:
Just let me know if you want me to submit a new PR instead, I can just create another branch and submit that and then we can continue the conversation there.

Rahix:
> Just let me know if you want me to submit a new PR instead, I can just create another branch and submit that and then we can continue the conversation there.

Recreating the branch didn't help on my end so I think that's the way to do ...

jkristell:
It was my fault, trying to be clever.

PR #55, opened 2020-10-04
Add atmega328pb

jkristell:


patch/atmega328pb.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - "0"
+TWI1:
+  _strip_end:
+    - "1"
+  TWCR:
+    _strip_end:
+      - "1"
+  TWSR:
+    _strip_end:
+      - "1"
```
Rahix, line 29: Re-adding from the other PR:

@Rahix:

> Hmm, I'm unsure about this ... This isn't the first time we're fighting with the odd naming scheme of AVR registers but I'm inclined to try sticking with the original names as much as possible.  In my opinion, it's better to have the macros from `avr-hal` flexible enough to take whatever names Atmel gave them than to try patching them all to fit.
> 
> The reason is simple: This way we have to deal with it once in the macro instead of everytime a new ATDF file with even more creative postfixes shows up.  Even worse, sometimes the numbers aren't at the end, but embedded somewhere in the middle of a register name.  Additionally, this makes cross-referencing C code with Rust easier as the names stay the same everywhere.
> 
> ---
> 
> I know the HAL currently isn't very consistent about this; `twi` makes the HAL crate spell out all register names entirely while `spi` doesn't expose the names at all currently.  I've recently updated at least the `usart` to a flexible macro that uses [`paste`](https://crates.io/crates/paste) (see [`avr-hal-generic/src/serial.rs`](https://github.com/Rahix/avr-hal/blob/master/avr-hal-generic/src/serial.rs)).  This seems to be working well so at some point `twi` and `spi` should probably follow suit.

@jkristell:

> I'm not sure either. The pros for doing it this way is that you don't have to write "specific but identical" code for the peripheral instances. In this special case the registers are suffixes with index, and the fields are too for the second device, but not the first. So you have to supply register field names as well :-/ and as you said, if the index is in the middle of the name you can't supply a empty suffix to the macro.
> 
> The only other devices I have worked with are stm32s, and they are doing the fixups in the "device" crate, so that's the primary reason I did it this way here. But I do think it's nice to fix the naming in some cases as they tend to be inconsistent for no really good reason and it just make reading and using them harder.

jkristell: Thinking some more about this and I think its better to do the renaming at the svd level.

Pros:
- More devices drivers can be shared at the hal-level
- Less macro magic in the hal
- The documentation gets clearer, no need to figure out if if spr1 and spr2 is different registers/fields or just different instances
- The suffix-naming is just C-ism we don't need it
- Common fixes like those for spi and twi can be applied more easily as shown in this PR
- register and field with suffixes "in the middle" can be updated with rename, so that's no problem

... for what it's worth. I mostly have been looking at the atmega328p and this device and it's been 10 years since I last worked with avrs :)

Rahix: I do see the benefits but I'm still unsure how much work it will be in the end.  I guess I'll have to play around with the new svd-patching a bit and see what is possible.

jkristell: I'm gonna do a version of the PR without the _strip_end calls and see how that works out with the hal.

Rahix: Hm ... I've been looking at the timer patches to see how we could remove the suffixes everywhere.  It got tedious very quickly, not sure if that effort is really worth it ...

My suggestion: Let's do it for SPI and I2C(TWI) for now and leave all other peripherals alone.  We can always come back and change things later.  Would you be okay with that?

jkristell: I have branches for the avr-device/hal that skips the register/field renaming (and just add a lots of * to the common patches) if you prefer that too. I can publish them as drafts tonight, so that you can compare and decide. I'm fine with either way for SPI and I2C.

Rahix (approved):
As stated in my comment above, I've decided that this current approach is fine for me.  Unless you object, can move move it out of draft status?

jkristell:
Will do! Sorry, been busy with other things.

Rahix:
No problem!  Is there anything you still wanted to do, or can I merge it as it is?  (We can always make changes in the future as well, if anything turns up, no stability guarantees for `avr-device` yet ;)

jkristell:
I think this can be merged as is and then I can revisit it when I have time to look at the Spi device

My plan was to do it all at in one session, but it's probably better to it like this, so that we get some progress :)

Rahix:
Sounds good.  Thanks for working on this!

PR #56, opened 2020-10-08
fix: attiny84 ADCSRA should be read-write

pitaj:
Haven't found any other variants which are affected by this, but it may be worth pulling all of the access-related patches into common or creating a common/tiny/adc patch.

### DIFF
--- a/patch/attiny84.yaml
+++ b/patch/attiny84.yaml
@@ -28,6 +28,9 @@ AC:
         bitWidth: 1
         access: read-write
 ADC:
+  _modify:
+    ADCSRA:
+      access: read-write
   ADCSRA:
     ADPS:
       _replace_enum:


Rahix:
Thanks for the PR!

cc @couchand, what do you think about this?  Should we (or even can we) extract a common patch here?

couchand:
I'm constantly amazed by the inconsistency in these ATDFs.

@pitaj Thanks for identifying this and taking care of it!

@Rahix I think the thing to do is remove the mux bits from patch/common/adc.yaml, since the prescaler enum seems consistent across devices and applying the read-write constraint even when it's not needed is harmless.  Then everything can use the common ADC patch, and the places it's being used now just need to define their mux variants.

couchand:
I'd be happy to take care of that refactor after this merges, or @pitaj if you feel like taking that on go ahead.  The usual way to verify a change like that is to back up the old patched directory and then diff it against the new one to make sure the patch refactor is clean.

pitaj:
Any idea when this fix will be released?

Rahix:
Right, next time nag me earlier about this!  I'll see whether I can get a few more things done today/tomorrow; if not I'll just do a release of the current state. :)

Rahix:
@pitaj, I released version 0.2.3 with this fix included just now :)

PR #57, opened 2020-10-12
fix: atmegaX8 TIFR should be read-write

no111u3:
Update AtmegaX8 Timers TIFR register to Write-Read (used in event driven code for clear overflow and other events without interrupt handler)


Rahix (approved):
Makes sense, thanks!

Issue #62, opened 2020-11-13
ATmega4809 support


explicite:
Looks like I need to spend more time on this chip. Require probably same changes as in https://github.com/Rahix/avr-device/issues/26#issuecomment-679831658

@Rahix  can you please advice which topic in [atdf2svd](https://github.com/Rahix/atdf2svd) I should progress 

```
Warning: Unhandled register node: "<register-group name=\"SINGLE\" offset=\"0\" name-in-module=\"TCA_SINGLE\" union-tag-value=\"0\" ...>"
Warning: Unhandled register node: "<register-group name-in-module=\"TCA_SPLIT\" name=\"SPLIT\" offset=\"0\" union-tag-value=\"1\" ...>"
Error: Register "DIR" has a weird name!
Error: Register "DIRCLR" has a weird name!
Error: Register "DIRSET" has a weird name!
Error: Register "DIRTGL" has a weird name!
Error: Register "IN" has a weird name!
Error: Register "INTFLAGS" has a weird name!
Error: Register "OUT" has a weird name!
Error: Register "OUTCLR" has a weird name!
Error: Register "OUTSET" has a weird name!
Error: Register "OUTTGL" has a weird name!
Error: Register "PIN0CTRL" has a weird name!
Error: Register "PIN1CTRL" has a weird name!
Error: Register "PIN2CTRL" has a weird name!
Error: Register "PIN3CTRL" has a weird name!
Error: Register "PIN4CTRL" has a weird name!
Error: Register "PIN5CTRL" has a weird name!
Error: Register "PIN6CTRL" has a weird name!
Error: Register "PIN7CTRL" has a weird name!
Error: Register "PORTCTRL" has a weird name!
Error: Register "DIR" has a weird name!
Error: Register "DIRCLR" has a weird name!
Error: Register "DIRSET" has a weird name!
Error: Register "DIRTGL" has a weird name!
Error: Register "IN" has a weird name!
Error: Register "INTFLAGS" has a weird name!
Error: Register "OUT" has a weird name!
Error: Register "OUTCLR" has a weird name!
Error: Register "OUTSET" has a weird name!
Error: Register "OUTTGL" has a weird name!
Error: Register "PIN0CTRL" has a weird name!
Error: Register "PIN1CTRL" has a weird name!
Error: Register "PIN2CTRL" has a weird name!
Error: Register "PIN3CTRL" has a weird name!
Error: Register "PIN4CTRL" has a weird name!
[...]

PR #63, opened 2020-11-18
ATmega4809 scaffold support #62 

explicite:
Scaffold for #62


svd/atmega4809.svd
```diff
@@ -0,0 +1,18300 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Rahix, line 1: The SVD files are not meant to be checked in.  That's why `svd/` is excluded in `.gitignore` :)

Rahix (changes requested):

Rahix:
For some reason, CI is failing right now ... I'm not quite sure why, maybe you can take a look and check whether this is also happening for you locally?

explicite:
```make: *** No rule to make target 'src/devices/atmega4809/mod.rs', needed by 'atmega4809'.  Stop.```
From CI and on local machine. 

Should I run svd2rust on svd and then copy mod.rs?

**Comment by Rahix** [comment]
Ah, I see.  I think the issue is that you must add a `patch/atmega4809.yaml` file for the Makefile to work right now ...

**Comment by explicite** [comment]
`patch/atmega4809.yaml` added. I was able to define ports from common. `make` pass. Still spot some problems with keaywords on rs

```
error: expected identifier, found keyword `async`
   --> src/devices/atmega4809/tcb0/ctrlb.rs:248:12
    |
248 |     pub fn async(&self) -> ASYNC_R {
    |            ^^^^^ expected identifier, found keyword
    |
help: you can escape reserved keywords to use them as identifiers
    |
248 |     pub fn r#async(&self) -> ASYNC_R {
    |            ^^^^^^^
```
Will try to dig into patching.

**Comment by Rahix** [comment]
Newer versions (built from git) of `svd2rust` automatically sanitize names that are keywords in rust.  But using those is currently blocked on Rahix/atdf2svd#13 ...

**Comment by explicite** [comment]
@Rahix `make` pass with common patch for `ports`. Should I fix something more in https://github.com/Rahix/atdf2svd?
Looks like we need to release it and then I can back to this and refine patches.

**Comment by Rahix** [comment]
I just released version [0.2.0](https://github.com/Rahix/atdf2svd/releases/tag/v0.2.0) of `atdf2svd` so you're good to go!  I recommend working with the git version of `svd2rust` as well

**Comment by explicite** [comment]
I started with AC. Still not sure what is the logic. @Rahix  do we have any guide for this? 
First diff which I spotted are modes for AC. From `read-write` we moved to `read-only`
Screenshot from 2020-11-26 22-09-24

**Comment by Rahix** [comment]
The right version is the patched one so the move was from `read-only` to `read-write` actually.

I'd take a look at the patches in `common/` and check if the changes that are done in each one also look necessary for `ATmega4809`.  If yes, include the patch.  Similarly for the timers, but you'll have to check which type of timers are used in your MCU.

Don't worry about patching everything from the get-go.  If all the common patches which apply are included, this is already a pretty good foundation and hopefully enough to make a HAL implementation work.

**Comment by explicite** [comment]
Only `common/ports.yaml` works without any changes. For rest I need to create custom patches. Should this be done with mcu suffix or in nested dir?

`common/usart.yaml`
```
svdtools.patch.MissingRegisterError: Could not find USART0:UCSR?A
```

`common/as.yaml`
```
svdtools.patch.MissingPeripheralError: Could not find AC
```

`common/adc.yaml`
```
svdtools.patch.MissingPeripheralError: Could not find ADC
```

`common/spi.yaml`
```
svdtools.patch.MissingRegisterError: Could not find SPI0:SPSR
```

`common/twi.yaml`
```
svdtools.patch.MissingRegisterError: Could not find TWI0:TWCR
```

`common/usart.yaml`
```
svdtools.patch.MissingRegisterError: Could not find USART0:UCSR?A
```

`common/wdt.yaml`
```
svdtools.patch.MissingRegisterError: Could not find WDT:WDTCSR,WDTCR
```

**Comment by Rahix** [comment]
hmm, I'd create a subdir `atmega_0x/`.

But it looks like the peripherals differ substantially from the ones we currently have on the other MCUs.  So it's probably not directly obvious what needs to be patches unless you want to go through the datasheet line by line ...

I'd actually keep patching to a minimun for now, and check back later, while doing the HAL implementation to see where changes are necessary.

File: patch/atmega4809.yaml [hunk 2]
```diff
@@ -0,0 +1,5 @@
+_svd: ../svd/atmega4809.svd
+
+_include:
+  - "common/port.yaml"
```
Rahix, line 4: Okay, maybe my previous comment was a bit all over the place.  To make it more clear what I meant:  Remove this patch as well, as it also does not really relate to the new AVR MCUs (the 0-series).

Then I will merge this PR and we can start looking into HAL driver implementations.  When things come up during HAL implementation, we can apply patches as needed.  I would not try to get everything right up front as that's not going to be possible anyway.

Rahix (changes requested):

patch/atmega4809.yaml
explicite, line 4:
```suggestion
```

patch/atmega4809.yaml
```diff
@@ -0,0 +1,3 @@
+_svd: ../svd/atmega4809.svd
+
+_include:
```
Rahix, line 3: CI is still failing, you have to remove the `_include:` as well:

```suggestion
```

Rahix (changes requested):

Rahix (approved):
Okay, remaining CI failures are because CI is using an old version of `svd2rust`.  I'm going to fix that later today, after that this branch should be ready to merge :)

Rahix:
Thanks a lot!

PR #64, opened 2020-11-19
Fix svd2rust regressions

Rahix:
Latest `svd2rust` built from git seems to be more strict on the validity of enumerated values (they must fit into the field they are for now).  This will break some of the current chip descriptions.

Fix all those cases so we are prepared for when `svd2rust` releases a new version.  A few of the changes will be breaking, which means the next version bump of `avr-device` after merging this PR must be a major one.

### DIFF
--- a/patch/atmega64.yaml
+++ b/patch/atmega64.yaml
@@ -6,3 +6,9 @@ _include:
   - "common/port.yaml"
   - "common/spi.yaml"
   - "common/usart.yaml"
+
+USART?:
+  UCSR?C:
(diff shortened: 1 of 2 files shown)


Rahix:
On a side note, as this kind of issue seems to come up again and again: Maybe `atdf2svd` should just not emit any broken enumerated values in the first place and drop them with a warning?

Rahix:
Dropped all changes which are no longer necessary when building with the latest version of `atdf2svd`.

PR #67, opened 2020-11-27
Add attiny841 and attiny861

jaxter184:
Preliminary pull request adding attiny841 and attiny861 chips to go with Rahix/avr-hal#105.


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "avr-device"
-version = "0.2.3"
+version = "0.2.4"
```
Rahix, line 3: Please drop this change.  I will do a version bump when it's time for a release; this should not be part of a pull-request.

Cargo.toml
```diff
@@ -34,6 +34,8 @@ atmega644 = []
 attiny84 = []
 attiny85 = []
 attiny88 = []
+attiny841 = []
+attiny861 = []
```
Rahix, line 38: I know it's a bit weird but please adhere to the alphabetical sorting that is present so far:

```diff
 attiny84 = []
+attiny841 = []
 attiny85 = []
+attiny861 = []
 attiny88 = []
```

Same goes for all the other changes as well.

patch/attiny841.yaml
```diff
@@ -0,0 +1,5 @@
+_svd: ../svd/attiny841.svd
+
+_include:
+  - "common/port.yaml"
```
Rahix, line 4: Are there really no other patches from `common/` which apply for this MCU?

patch/attiny861.yaml
```diff
@@ -0,0 +1,5 @@
+_svd: ../svd/attiny861.svd
+
+_include:
+  - "common/port.yaml"
```
Rahix, line 4: Same here?

Rahix (changes requested):
Thanks for your PR!  I've got some minor comments, overall the changeset looks good!

Cargo.toml
jaxter184: Oops, I totally intended to do that, but for some reason my brain said that 41 is bigger than 8. My bad

patch/attiny841.yaml
jaxter184: I'll look into this some more

jaxter184:
I still need to figure out the timer stuff, and I probably want to look through the patches a little more closely to make sure that I'm doing it correctly

src/lib.rs
```diff
@@ -25,7 +27,9 @@
 //! * `atmega32u4`
 //! * `atmega48p`
 //! * `atmega64`
+//! * `attiny841`
```
trembel, line 30: Here the `attiny84` is missing (already on the attiny84's original commit).

jaxter184: It seems like atmega644 is also missing

jaxter184:
I think I've just about figured out the timer stuff, with one caveat: In `patch/timer/dev`, there are a 2 different 16-bit timer patches. The primary difference that I could distinguish is that one describes that `OCix` is set/cleared at `TOP` rather than `BOTTOM`. However, the datasheet register descriptions are the same between the devices that use `16bit-tiny84-tc1.yaml` (attiny84 table 14-3) and the devices that use `16bit.yaml` (for example: atmega2560 table 20-3), namely that both say to set and clear at `BOTTOM`. Am I reading these wrong? Is there an error? Should I be looking at a different part of the datasheet for this information?

src/lib.rs
Rahix: Ah, I'll add those after merging this PR.

Rahix:
> I think I've just about figured out the timer stuff, with one caveat: In patch/timer/dev, there are a 2 different 16-bit timer patches. The primary difference that I could distinguish is that one describes that OCix is set/cleared at TOP rather than BOTTOM. However, the datasheet register descriptions are the same between the devices that use 16bit-tiny84-tc1.yaml (attiny84 table 14-3) and the devices that use 16bit.yaml (for example: atmega2560 table 20-3), namely that both say to set and clear at BOTTOM. Am I reading these wrong? Is there an error? Should I be looking at a different part of the datasheet for this information?

Looks like this slipped through and you're right, they are set/cleared at `BOTTOM`.  The other difference between those files is the `TIFRn` register.  Maybe that doesn't exist on the ATtiny version?

patch/attiny841.yaml
```diff
@@ -0,0 +1,66 @@
...
+      ACIS?:
+        description: "Analog Comparator Interrupt Mode Select"
+      ACO:
+        access: read-only
+    ACIS?:
+      _replace_enum:
+        ON_TOGGLE: [0, "Interrupt on Toggle"]
+        # Leaving [1, 'Reserved'] out
+        ON_FALLING_EDGE: [2, "Interrupt on Falling Edge"]
+        ON_RISING_EDGE: [3, "Interrupt on Rising Edge"]
```
Rahix, line 26: You can make the common patch work here as well by adding the following changes:

```diff
 AC:
   _modify:
-    ACSR:
+    ACSR,ACSR?A:
       access: read-write
-  ACSR:
+  ACSR,ACSR?A:
     _modify:
-      ACIS:
+      ACIS,ACIS?:
         description: "Analog Comparator Interrupt Mode Select"
       ACO:
         access: read-only
-    ACIS:
+    ACIS,ACIS?:
       _replace_enum:
         ON_TOGGLE: [0, "Interrupt on Toggle"]
         # Leaving [1, 'Reserved'] out

```

patch/attiny861.yaml
```diff
@@ -0,0 +1,26 @@
...
+      ACIS:
+        description: "Analog Comparator Interrupt Mode Select"
+      ACO:
+        access: read-only
+    ACIS:
+      _replace_enum:
+        ON_TOGGLE: [0, "Interrupt on Toggle"]
+        # Leaving [1, 'Reserved'] out
+        ON_FALLING_EDGE: [2, "Interrupt on Falling Edge"]
+        ON_RISING_EDGE: [3, "Interrupt on Rising Edge"]
```
Rahix, line 26: Similar change to above should work here as well I think.

patch/timer/attiny861.yaml
```diff
@@ -0,0 +1,16 @@
...
+  TC0:
+    description: "Timer/Counter0"
+  TC1:
+    description: "Timer/Counter1, 10-bit"
+
+TC0:
+  _include:
+    - "dev/16bit-tiny861-tc0.yaml"
+
+TC1,TC2:
```
Rahix, line 14: No `TC2`, right?

Rahix (changes requested):

jaxter184:
> > I think I've just about figured out the timer stuff, with one caveat: In patch/timer/dev, there are a 2 different 16-bit timer patches. The primary difference that I could distinguish is that one describes that OCix is set/cleared at TOP rather than BOTTOM. However, the datasheet register descriptions are the same between the devices that use 16bit-tiny84-tc1.yaml (attiny84 table 14-3) and the devices that use 16bit.yaml (for example: atmega2560 table 20-3), namely that both say to set and clear at BOTTOM. Am I reading these wrong? Is there an error? Should I be looking at a different part of the datasheet for this information?
> 
> Looks like this slipped through and you're right, they are set/cleared at `BOTTOM`. The other difference between those files is the `TIFRn` register. Maybe that doesn't exist on the ATtiny version?

It looks like `TIFRn` does exist in the ATtiny84. Should I maybe consolidate it into one file? Is it bad practice to put multiple unrelated changes in a single pull request?

patch/timer/attiny861.yaml
jaxter184: whoops

patch/attiny861.yaml
jaxter184: When I added this change, it broke the AC patch for atmega328pb, which I think is because of an error in the svd file for that chip. According to the svd file, there's a register called ACSRA with no fields (in addition to the ACSR register with the ACIS field), but the ACSRA field doesn't show up at all in the datasheet as far as I can tell. For now, I made the changes to the common AC patch, and replaced the common patch in atmega328pb.yaml with one that works. I'm not sure this is the best solution though.

Rahix:
> It looks like TIFRn does exist in the ATtiny84. Should I maybe consolidate it into one file?

If possible, that would of course be the best!

> Is it bad practice to put multiple unrelated changes in a single pull request?

I guess this is highly project specific.  Here's my take:  I think it is very important that every "atomic" change goes into a separate _commit_.  I.e. every change that makes "sense" on its own and can stand without the following changes being needed should be separated.

The reason for this is that this means the git history is a clean record of what was done with the project.  If I find a bug in a few years, I can just [`git blame`](https://www.git-scm.com/docs/git-blame) and find out _what_ was changed and ideally also _why_ (if a proper commit message was written).  Similarly, if something stops working at some point I can find the exact change that introduced a regression using [`git bisect`](https://www.git-scm.com/docs/git-bisect).

If multiple changes happen in a single commit or if an "atomic" change is split out over multiple commits, this is of course not going to work effectively.

Now, to archieve this clean separation, there are two ways:

- If contributors are willing to take care of this, they can rebase their changeset to split it into such clean, isolated commits (and document each one with a good commit message).  This would be super helpful for me, because I can then just rebase the commit series and all is good.
- If contributors send a PR with more or less random commits or non-descriptive commit messages, I'll have to squash the series and write a proper message explaining all changes.  This of course means that potentially multiple changes are merged into a single commit ...

I see that you've already followed the first route which is really nice (thanks a lot!).  So I have no problem if you add more changes into this series.  A minor nag if you want to make it even more perfect: You could reorder the commits slightly so that each one could possibly compile (right now the ATmega328PB fix is after the changes to the AC common patch).  But that's really a minor detail.

---

On a side note: There is currently a merge conflict.  Please rebase on master to fix :)

jaxter184:
I've `rebase`d the commits, and assuming my testing procedure was correct, they are all `make`able.

patch/timer/dev/16bit-tiny861-tc0.yaml
```diff
@@ -1,13 +1,6 @@
...
-      _write_constraint: enum
-  COM??:
-    _replace_enum:
-        DISCONNECTED:  [0, "Normal port operation, OCix disconnected"]
-        MATCH_TOGGLE:  [1, "Toggle OCix on Compare Match (Might depend on WGM)"]
-        MATCH_CLEAR:   [2, "Clear OCix on Compare Match (If PWM is enabled, OCix is set at BOTTOM)"]
-        MATCH_SET:     [3, "Set OCix on Compare Match (If PWM is enabled, OCix is cleared at BOTTOM)"]
+_modify:
+  TIFR?:
+    access: read-write
```
Rahix, line 3: Just for my understanding: The tiny861 TC0 doesn't have any `TCCR#A[COM##]` fields and thus needs a separate patch.  Is that correct?

Rahix (approved):
Looks really good and thanks for taking the time to properly split out the commits!

patch/timer/dev/16bit-tiny861-tc0.yaml
jaxter184: The t861 16-bit timer is a little weird compared to some of the other 16-bit timers, and I think this is because it can be split into 2 8-bit timers (with some caveats). I haven't come across this in any other chips, so should I maybe have put it into `patch/timer/attiny861.yaml` instead of a separate timer file, kinda like how it's done it `patch/timer/atmega328pb.yaml`? If so, I think it's something that could be done in a separate commit, since the tiny88 and tiny85 also have separate timer files.

Either way, I think the answer to your question is yes, `TCCR0A` does not contain `COM##` fields, or any other multi-bit fields that are enumerated.

Rahix:
Thanks for you work in this PR!

PR #69, opened 2020-12-03
Replace `gen-intr-lut.sh` hack with `svdtools` solution

explicite:
@Rahix not sure if I made correct change to [Makefile](https://github.com/Rahix/avr-device/pull/69/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52)
Please review. Thanks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,9 +51,9 @@ src/devices/%/mod.rs: src/devices/%/mod.full.rs
 	@# to the global version
 	@patch --no-backup-if-mismatch --quiet $@ patch/modrs.patch
 	@echo -e "\tGEN-VECTOR\t>macros/src/vector.rs"
-	@./gen-intr-lut.sh src/devices/*/interrupt.rs >macros/src/vector.rs
+	@./gen-intr-lut.sh svd/*.patched >macros/src/vector.rs
 
-macros/src/vector.rs: src/devices/*/interrupt.rs
+macros/src/vector.rs: svd/*.patched
 	@echo -e "\tGEN-VECTOR\t>macros/src/vector.rs"
 	@./gen-intr-lut.sh $^ >$@
 
--- a/Makefile
+++ b/Makefile
@@ -50,9 +50,11 @@ src/devices/%/mod.rs: src/devices/%/mod.full.rs
 	@# Remove DEVICE_PERIPHERALS declaration and replace it with a reference
 	@# to the global version
 	@patch --no-backup-if-mismatch --quiet $@ patch/modrs.patch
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -50,11 +50,9 @@ src/devices/%/mod.rs: src/devices/%/mod.full.rs
 	@# Remove DEVICE_PERIPHERALS declaration and replace it with a reference
 	@# to the global version
 	@patch --no-backup-if-mismatch --quiet $@ patch/modrs.patch
-	@echo -e "\tGEN-VECTOR\t>macros/src/vector.rs"
```
Rahix, line 53: Please don't remove the `@echo` lines here and below.

gen-intr-lut.sh
```diff
@@ -7,12 +7,10 @@ echo     "pub fn lookup_vector(chip: &str, intr: &str) -> Option<usize> {"
 echo     "    match chip {"
 
 for intr_path in "$@"; do
-    chip="$(basename "$(dirname "$intr_path")")"
+    chip="$(basename "${intr_path%%.*}")"
```
Rahix, line 10: This is a bit non-obvious.  I'd rather make it explicit what we're removing:

```suggestion
    chip="$(basename "${intr_path%.svd.patched}")"
```

gen-intr-lut.sh
```diff
@@ -7,12 +7,10 @@ echo     "pub fn lookup_vector(chip: &str, intr: &str) -> Option<usize> {"
...
     echo "        \"$chip\" => match intr {"
 
-    sed '/=> Ok(Interrupt::.\+),$/!d
-    s/ \+\(.\+\) => Ok(Interrupt::\(.\+\)),$/            "\2" => Some(\1),/' "$intr_path"
-
+    svd interrupts --no-gaps $intr_path | awk '{print "            \""toupper(substr($2, 1, length($2)-1))"\"" " => Some(" $1"),"}'
```
Rahix, line 13: I think it would make sense to add a comment on why `toupper()` is used here:  Because `svd2rust` internally [also touppers()](https://github.com/rust-embedded/svd2rust/blob/56be78729279eeebef65110c13be8d96c0b9270f/src/generate/interrupt.rs#L44) the interrupt names (was news to me!).

Rahix (changes requested):
Thanks for your PR!  I've left a few minor comments, but I gave it a test and everything looks good!

Makefile
explicite: reverted

Rahix (approved):
Looks good, thanks!

PR #74, opened 2021-02-05
Split INT field into 2 bits for atmega328 and atmega328pb

gpgreen:
The INT0 and INT1 fields in the EIMSK register were incorrectly
combined into a 2 bit field, split into individual bits
Allowed write access to PRR0 and PRR1 in atmega328pb

### DIFF
--- a/patch/atmega328p.yaml
+++ b/patch/atmega328p.yaml
@@ -15,3 +15,10 @@ CPU:
   _modify:
     PRR:
       access: read-write
+
+EXINT:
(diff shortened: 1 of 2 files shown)


gpgreen:
I downloaded the atmega328pb datasheet and compared with the generated svd. It had the same problems with the EIMSK and PRR registers (multiple in this case). So I went ahead and added those also.

Rahix (approved):
Looks good, thanks!

PR #77, opened 2021-03-08
add atmega8u2

drmorr0:
## Testing done

* I've confirmed that the source for the Atmega8u2 builds and compiles.
* I've linked in this version of `avr-device` into `avr-hal` and confirmed that it compiles (https://github.com/Rahix/avr-hal/pull/159)

@Rahix PTAL :D


Rahix (approved):
Looks good to me, thanks for working on this!

PR #79, opened 2021-04-13
implement cortex-m-rt-macros static mut conversion

tones111:
This change brings in the cortex-m-rt macro support to identify and convert static mut variables as described in the embedded rust book chapter on [exceptions](https://docs.rust-embedded.org/book/start/exceptions.html).

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -93,6 +93,13 @@ pub fn entry(
         .collect::<Vec<_>>();
 
     quote::quote! (
+        #[cfg(not(any(doc, target_arch = "avr")))]
+        compile_error!(
+            "Ensure that you are using an AVR target! You may need to change \
+       directories or pass a --target flag to cargo. See
+       https://github.com/Rahix/avr-device/pull/41 for more details."
+        );
+
         #[doc(hidden)]
         #[export_name = "main"]
         pub unsafe extern "C" fn #tramp_ident() {
@@ -101,6 +108,7 @@ pub fn entry(
             )
         }
 
+        #[doc(hidden)]
         #f
     )
     .into()
@@ -222,6 +230,7 @@ pub fn interrupt(
             )
         }
 
+        #[doc(hidden)]
         #f
     )
     .into()
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -11,7 +11,7 @@ pub fn entry(
     args: proc_macro::TokenStream,
     input: proc_macro::TokenStream,
 ) -> proc_macro::TokenStream {
-    let f = syn::parse_macro_input!(input as syn::ItemFn);
+    let mut f = syn::parse_macro_input!(input as syn::ItemFn);
 
     // check the function signature
     let valid_signature = f.sig.constness.is_none()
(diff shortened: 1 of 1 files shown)


Rahix:
Nice work, thanks a lot!  I'll take a closer look and try this out later.  The CI failures are something I need to fix, it's using an older version of rustc.  You'll need to rebase once I have fixed it.

Rahix:
Hey @tones111,

super sorry for not reacting until now :(  I've pushed a fix for the CI issue to `main`, can you please rebase you branch ontop of that so CI in this PR will start working again?

macros/src/lib.rs
```diff
@@ -47,31 +44,64 @@ pub fn entry(
...
+        })
+        .collect::<Vec<_>>();
 
     quote::quote! (
-        #[cfg(not(any(doc, target_arch = "avr")))]
-        compile_error!(
-            "Ensure that you are using an AVR target! You may need to change \
-       directories or pass a --target flag to cargo. See
-       https://github.com/Rahix/avr-device/pull/41 for more details."
-        );
```
Rahix, line 67: Please don't drop this safeguard error.

macros/src/lib.rs
```diff
@@ -47,31 +44,64 @@ pub fn entry(
...
         #[export_name = "main"]
-        pub #unsafety extern "C" fn #ident() -> ! {
-            #(#stmts)*
+        pub unsafe extern "C" fn #tramp_ident() {
+            #ident(
+                #(#resource_args),*
+            )
         }
+
+        #f
```
Rahix, line 104: Previously we had the `#[doc(hidden)]` attribute on the main function to hide it under all circumstances.  I see you've left it on the trampoline, but shouldn't we also add it to the function itself?  Especially because we change its name to `__avr_device_rt_main` I don't see any reason not to...

macros/src/lib.rs
```diff
@@ -149,18 +211,76 @@ pub fn interrupt(
...
+                #(#resource_args),*
+            )
         }
 
-        #(#attrs)*
-        #[allow(non_snake_case)]
-        #unsafety fn #ident() {
-            #(#stmts)*
-        }
+        #f
```
Rahix, line 225: Same here, I think the vector function should be annotated with `#[doc(hidden)]`.

Rahix (changes requested):
Thanks for the quick response, I've now finally found some time to review your PR and give it a try.  The generated code looks good, there are just a few minor details about which I've left comments below.

Thanks again for your work on this and sorry for my slow responses :/

macros/src/lib.rs
tones111: Ah, right... doesn't make any sense to remove this check.  I've restored it and added the requested '''#[doc(hidden)]''' annotations.  Thanks for the review!

Rahix (approved):
Looks great, thanks for your work on this (and for your patience)!  I want to run a full test before merging, hopefully later today.  But I strongly assume that everything will work, so I'm going to approve the PR already :)

Rahix:
Test went fine, thanks a lot!

Issue #26, opened 2020-06-16
attiny412 support


trembel:
Hi
I would like to have support for the attiny412. As it is a MCU of the new AVR-1 series, will it work with your tool?

Also am I understanding correctly, that this crate is providing something similar to the "PAC" crates for the different cortex-m MCUs?

Best,
trembel

Rahix:
Yes, `avr-device` is the PAC crate with the raw register definitions.  [`avr-hal`](https://github.com/Rahix/avr-hal) builds on top and has safe abstractions over the peripherals.

> As it is a MCU of the new AVR-1 series, will it work with your tool?

I don't see a problem with that, it should work just fine.

Are you interested in taking a stab at adding the support yourself?  I can guide you through it if you want.

Word of caution if you don't yet have the avr-rust toolchain set up:  There's quite a few pitfalls still, you need the `avr-support` branch from https://github.com/avr-rust/rust as the recently merged upstream support still has a few issues for now.

trembel:
Perfect, thank you.

At the moment I'm to busy, but I'll give it a try in a month. I think your "instruction" is quite self-explaining, but if I get trouble I'll ask you :)

Rahix:
Sounds good, thanks a lot! :)

trembel:
I'm at the moment trying to include the attiny412. However, on compiling atdf2svd fails:
```
Warning: Unhandled register node: "<register-group offset=\"0\" union-tag-value=\"0\" name=\"SINGLE\" name-in-module=\"TCA_SINGLE\" ...>"
Warning: Unhandled register node: "<register-group name=\"SPLIT\" offset=\"0\" union-tag-value=\"1\" name-in-module=\"TCA_SPLIT\" ...>"
Error: Register "DIR" has a weird name!
Error: Register "DIRCLR" has a weird name!
Error: Register "DIRSET" has a weird name!
Error: Register "DIRTGL" has a weird name!
Error: Register "IN" has a weird name!
Error: Register "INTFLAGS" has a weird name!
Error: Register "OUT" has a weird name!
Error: Register "OUTCLR" has a weird name!
Error: Register "OUTSET" has a weird name!
Error: Register "OUTTGL" has a weird name!
Error: Register "PIN0CTRL" has a weird name!
Error: Register "PIN1CTRL" has a weird name!
Error: Register "PIN2CTRL" has a weird name!
Error: Register "PIN3CTRL" has a weird name!
Error: Register "PIN4CTRL" has a weird name!
[...]

Issue #66, opened 2020-11-26
Escaping/Renaming `ASYNC` register


trembel:
Trying to compile for `attiny412` following error occurs, due to the naming of the `ASYNC` register in the `TCB0` and `TCD0` peripheral:
```
error: expected identifier, found keyword `async`
   --> src/devices/attiny412/tcb0/ctrlb.rs:248:12
    |
248 |     pub fn async(&self) -> ASYNC_R {
    |            ^^^^^ expected identifier, found keyword
    |
help: you can escape reserved keywords to use them as identifiers
    |
248 |     pub fn r#async(&self) -> ASYNC_R {
    |            ^^^^^^^

error: expected identifier, found keyword `async`
   --> src/devices/attiny412/tcb0/ctrlb.rs:270:12
    |
270 |     pub fn async(&mut self) -> ASYNC_W {
    |            ^^^^^ expected identifier, found keyword
    |
help: you can escape reserved keywords to use them as identifiers
    |
270 |     pub fn r#async(&mut self) -> ASYNC_W {
    |            ^^^^^^^

error: expected identifier, found keyword `async`
   --> src/devices/attiny412/tcd0/evctrla.rs:256:12
    |
256 |     pub fn async(self) -> &'a mut W {
    |            ^^^^^ expected identifier, found keyword
    |
help: you can escape reserved keywords to use them as identifiers
    |
256 |     pub fn r#async(self) -> &'a mut W {
    |            ^^^^^^^

error: expected identifier, found keyword `async`
   --> src/devices/attiny412/tcd0/evctrlb.rs:256:12
    |
256 |     pub fn async(self) -> &'a mut W {
    |            ^^^^^ expected identifier, found keyword
    |
help: you can escape reserved keywords to use them as identifiers
    |
256 |     pub fn r#async(self) -> &'a mut W {
    |            ^^^^^^^

```

Not sure if this should be fixed in `atdf2svd` by renaming `ASYNC` registers, or here.

Rahix:
Try using a version of `svd2rust` built from git, I believe this is fixed over there.

trembel:
Yes, this is fixed - but introduces a new error (for **ALL** avr devices):
```
error[E0433]: failed to resolve: use of undeclared crate or module `cortex_m`
  --> src/devices/attiny412/interrupt.rs:52:13
   |
52 | unsafe impl cortex_m::interrupt::InterruptNumber for Interrupt {
[...]

PR #82, opened 2021-07-17
Add ATtiny202 support

anall:
I am not sure if this actually produces valid register accesses for an ATtiny202 as my chips will not show up until August -- but my firmware does compile.

All of the svd patches are in `patch/common/attiny-0-series.yaml` which should work for all of the 0-series (but may only work for ATtiny20* and ATtiny40*, but not 80* or 160* given how the datasheet is split).

The yaml patch might even be somewhat applicable to the 1-series? They do seem _extremely_ similar through my brief research.

I am currently only using the TWI/I²C peripheral and the PORT registers so have not really looked at the other peripherals for patching besides a few random things that stood out to me. I tried to look at the other patch files for ideas, but the 0-series seems significantly different. If you have any suggestions for things I should focus on I would be happy to take a look.

I can extract and commit the `atdf` files for the rest of the files that the ATtiny202 datasheet covers if you would like.

This may be tangentially related to #26 (if anything else, assuming the 1-series `atdf` files are similar, it seems the workflow works now)


patch/common/attiny-0-series.yaml
```diff
@@ -0,0 +1,57 @@
+# This is probably good for at least ATtiny202/204/402/404/406, but may not be valid for ATtiny80* and ATtiny160*
+
+TCB0:
+  CTRLB:
+    _modify:
+      # 'async' is reserved
+      ASYNC:
+        name: "TCB_ASYNC"
```
anall, line 8: This is the only change that is hard _required_, as `async` is a reserved keyword.

Rahix: Technically it shouldn't be because newer versions of `svd2rust` will use a raw literal for such registers (`r#async`).  But I'm fine with changing it either way.

Rahix (approved):
Thanks for the PR, looks good to me from my standpoint.  I'd be willing to merge it like it is - if anything comes up when you test this against actual hardware, we can always come back and update it later.

> I can extract and commit the atdf files for the rest of the files that the ATtiny202 datasheet covers if you would like.

Unless there are devices that you're actually intending to use, I'd prefer not to.  `avr-device` is getting quite bloated already so I'd want to only add devices which have actual users.  That way I can push the date where we need to split up this crate into many down further into the future ;)

patch/common/attiny-0-series.yaml
Rahix: (was previously discussed in #66 as well)

anall: Huh, I could have sworn I installed the newest version of `svd2rust` when I started on this, but it seems like I was on 0.17.

Although upgrading `svd2rust` (to either 0.18 or 0.19) fails (see #83)

Rahix: I see, I'll need to adjust the buildsystem to the new version...  But let's not block this PR on that - CI tested with a working revision of svd2rust and is happy.

PR #84, opened 2021-07-27
Add ATtiny167 support

quentinmit:
(as used on the Digispark Pro board)

This builds for me, but is completely untested on real hardware. Undoubtedly there will be further patches needed but I think this might be mergeable as-is for people to get started poking at it.


Rahix (approved):
Looks good to me, thanks a lot!

PR #85, opened 2021-08-27
Add support for ATtiny2313/A

USA-RedDragon:
This is for both the [ATtiny2313](http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-2543-AVR-ATtiny2313_Datasheet.pdf) and the [ATtiny2313a](https://ww1.microchip.com/downloads/en/DeviceDoc/doc8246.pdf). They're pin-compatible, but some registers and electrical properties are changed between them.


Rahix (changes requested):
Hi,

thank you very much for the PR!  Everything is looking quite good already, however it seems there are still some errors.  With svd2rust 0.19 (for which support was just merged to `main`) I get:

```text
[ERROR svd2rust] In device `ATtiny2313`
    
    Caused by:
        0: In peripheral `USART`
        1: In register `UCSRC`
        2: Parsing field #4
        3: In field `UMSEL`
        4: Value 3 out of range 0..2
```

I suggest rebasing your work on latest `main` and retry building to verify this locally.  I think the error might just be that on ATtiny2313, `UMSEL` can't be `SPI_MASTER` because the field is just 1 instead of 2 bits.  But maybe verify this against the datasheet first...

USA-RedDragon:
Ah you're exactly right: the original non-A ATtiny2313's UCSRC UMSEL is only 1 but. I'll get a fix added. 

Thanks

USA-RedDragon:
Screenshot_20210828-133952_Drive

Documenting for posterity

Rahix (approved):
Looks good now, thanks a lot!

Issue #88, opened 2021-10-02
`critical-section` support


torrancew:
In attempting to integrate the `avr-hal` ecosystem with `embedded-nal`, I ran into a series of hurdles involving `heapless`,`atomic-polyfill` and `critical-section`.

Ultimately, if `avr-device` were to implement the API provided by [embassy-rs/critical-section](https://github.com/embassy-rs/critical-section/), integrating with `embedded-nal` and its dependencies would be greatly simplified.

Would you be open to a PR that makes the required (minor) changes to `avr_device::interrupt` to accomodate this? I can see two main ways to go about it:

## Option 1: Add support for AVR targets to `critical-section` by having it depend upon `avr_device` when building for the AVR platform
The main missing piece here would be a public API that `critical-section` can use to read the GIE field of `SREG`. Since this register is not exposed by `avr_device`, we would need to surface it in some way or another (either a public method that invokes the underlying LLVM assembler, or exposing it in a way more consistent with the SVD-backed registers. This would open the possibility of migrating the `avr_device::interrupt::free` API to the `critical_section::with` API, which is equivalent in signature, but not in name.

Side note: I'm unsure on the rationale for patching `SREG` out of SVDs. Is it simply a matter of not wanting the exposure to be a `safe` API?

## Option 2: Implement `critical-section` via its `custom-impl` feature, entirely in `avr_device`
I believe this would be less ergonomic for users who inadvertently pull in both, but could be mistaken here, as I haven't had a lot of exposure to the nuances of Cargo features & how they propagate in complex dependency graphs. The only change to the `avr-device` API in this case would be the net-new implementation of the trait provided by `critical-section` (which closely mirrors the existing interrupt API).

In either case I'm more than willing to write the patches, both to `avr_device` and to `critical-section`, but wanted to see if there was a preferred direction on your end.

Rahix:
Hi,

yes, I think adding the necessary API to support `critical-section` is a good idea and "Option 1" is certainly the better way forward.  Adding a helper to check the status of the interrupt bit is not a problem and might be beneficial beyond this usecase.

> Side note: I'm unsure on the rationale for patching SREG out of SVDs. Is it simply a matter of not wanting the exposure to be a safe API?

[...]

PR #89, opened 2021-10-03
Return the value of `SREG` upon disabling interrupts

torrancew:
This commit addresses #88 by returning the original value of `SREG` upon
disabling interrupts. This allows `critical-section` to implement itself
for AVR, while not breaking the existing `avr-device` API.

The atlernative of exposing a function to return `SREG` as part of the
API was considered, but it was not clear where best to fit such a
function in the structure of the crate, and overall, this approach
seemed a bit more elegant.


torrancew:
Edited the above to fix a minor spelling error

Rahix (commented):
Hi, thanks for the PR!

I like the general approach but I think we should only return the interrupt bit, not the entire SREG.  This would mean to make `disable()` return a `bool` which is `SREG & 0x80 != 0x00` IIRC.

torrancew:
@Rahix I've updated the commit message and code in line with what you suggested. Let me know if there are any other changes you'd like to see, and I'll get those in ASAP.

PR #93, opened 2022-03-06
Add ATmega128RFA1 support

the6p4c:
Adds basic ATmega128RFA1 support.

This has been partially tested (GPIO, SPI) with changes I've got pending to `avr-hal` libraries, but I don't have a good enough development board to test the other peripherals more than ensuring they build.

This PR also adjusts the common USART patches for `UCSZ?` slightly to work with the ATmega128RFA1's ATDF. The ATDF does specify enum values for the field but the patch was not `_replace_enum`. I've changed the common patch to use `_replace_enum`, to make sure the fields are consistent with other micros.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ all-features = true
 device-selected = []
 at90usb1286 = ["device-selected"]
 atmega1280 = ["device-selected"]
+atmega128rfa1 = ["device-selected"]
 atmega168 = ["device-selected"]
 atmega2560 = ["device-selected"]
 atmega8 = ["device-selected"]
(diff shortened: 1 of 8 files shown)


Rahix (approved):
Looks good, thanks for the PR!  Unless you want to approach it differently, I would merge it like it is right now and if anything comes up later during HAL integration, it can be fixed in follow-up pull requests.

the6p4c:
Sounds good to me. HAL integration seems fine with those peripherals (GPIO/SPI) so I'd like to say I'm confident with the rest, but fixing when those comes up seems smart :D

PR #94, opened 2022-03-11
Add support for ATmega1284p

mguillemot:
Title says it all :)

I used the atmega328p as inspiration for the SVD patches since it's fairly similar to atmega1284p. I did apply a few more splits for registers in EXINT to make handling interrupts a bit smoother.


Rahix (approved):
Hey!

Changeset looks good, thank you very much!

PR #99, opened 2022-06-04
Use the same timer patch for atmega1280 as for atmega2560

dalpil:
It seems like the timer patch for the atmega2560 is applicable for the atmega1280 as well, unless I'm mistaken.

I stumbled over the lack of timer patches for the atmega1280 while working on https://github.com/Rahix/avr-hal/pull/272 , this PR should fix this by using the same patch for the 1280 as for the 2560.

As far as I can tell the only difference between the two MCU's seems to be the amount of flash memory available(?).


Rahix (approved):
Hey, this looks good, thank you very much!

PR #100, opened 2022-06-23
Add support for attiny13a

alec-deason:
This adds configurations for the attiny13a. I followed the example of the attiny85 which is a very similar chip. I haven't pushed very hard on this but have been able to use it to write some code that uses the 13a's ADC.


Rahix (approved):
Thanks a lot!

PR #101, opened 2022-06-24
Add support for ATmega164pa

tronje:
Adds the ATDF file provided from Microchip, and makes changes analog to
290613454fbdc5e4ac98e53deccaf74dafc88963.
This has been tested on real hardware.

We've got some changes to avr-hal in the works as well, but the commits there need some cleaning up. We'll fix that after this has been merged.

Let us know what you think :)


Rahix (approved):
Looks good, thank you!

PR #102, opened 2022-07-06
Adds patches for the timers on attiny167

dalpil:
Adds patches for TC0 and TC1 on the attiny167 so that prescaler values (and the `COMnx` register values) can be set by human readable methods (`.prescale_64()` instead of `.val_0x03()` for example).


Rahix (approved):
Perfect, thank you!

PR #103, opened 2022-08-03
Add basic documentation for beginners

mbuesch:
There are many different traits and structs that interoperate in very complex ways in this crate. That makes it very hard to understand the basic concept of the crate. It's hard to find out where to look first in the documentation.
As always, examples help a lot when explaining things.

This pull request adds a couple of basic examples to help kick starting the knowledge of people new to this project.

### DIFF
--- a/src/interrupt.rs
+++ b/src/interrupt.rs
@@ -3,9 +3,29 @@
 //! For the most part, [crate::interrupt::free] is what you want:
 //!
 //! ```
-//! atmega32u4::interrupt::free(|cs| {
+//! avr_device::interrupt::free(|cs| {
 //!     // Interrupts are disabled here
-//! })
+//! });
+//! ```
+//!
+//! To access shared state, Mutex can be used:
+//!
+//! ```
+//! use avr_device::interrupt::Mutex;
+//! use core::cell::Cell;
+//!
+//! // Use Cell, if the wrapped type is Copy.
+//! // Use RefCell, if the wrapped type is not Copy or if you need a reference to it for other reasons.
+//! static MYGLOBAL: Mutex<Cell<u16>> = Mutex::new(Cell::new(0));
+//!
+//! avr_device::interrupt::free(|cs| {
+//!     // Interrupts are disabled here
+//!
+//!     // Acquire mutex to global variable.
+//!     let myglobal_ref = MYGLOBAL.borrow(&cs);
+//!     // Write to the global variable.
+//!     myglobal_ref.set(42);
+//! });
 //! ```
 
 pub use bare_metal::{CriticalSection, Mutex, Nr};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,34 +35,101 @@
 )]
 //! Which chips the crate is built for depends on the feature flag used.
 //! The following chips are available (using feature flags of the same name):
-//! * `at90usb1286`
-//! * `atmega1280`
-//! * `atmega1284p`
-//! * `atmega128rfa1`
-//! * `atmega164pa`
-//! * `atmega168`
-//! * `atmega2560`
-//! * `atmega8`
-//! * `atmega8u2`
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -63,6 +63,59 @@
...
+//! ## Main program entry point
+//!
+//! ```
+//! #[avr_device::entry]
+//! fn main() -> ! {
+//!     loop {
+//!         // Your code here.
+//!     }
+//! }
+//! ```
```
Rahix, line 78: I think we also need to mention here that the `rt` crate feature must be active for this.

src/interrupt.rs
```diff
@@ -3,9 +3,27 @@
...
+//! });
+//! ```
+//!
+//! To access shared state, Mutex can be used:
+//!
+//! ```
+//! use avr_device::interrupt::Mutex;
+//! use core::cell::Cell;
+//!
+//! static MYGLOBAL: Mutex<Cell<u16>> = Mutex::new(Cell::new(0));
```
Rahix, line 17: I think it would be valuable to point out that `Cell` should be used for `Copy` types and `RefCell` when a reference to the data inside the mutex is needed at some point.

src/lib.rs
```diff
@@ -63,6 +63,59 @@
 //! * `attiny85`
 //! * `attiny861`
 //! * `attiny88`
+//!
+//! # How to use this crate?
```
Rahix, line 67: Maybe mention that this is the low-level peripheral access crate (PAC) and there are higher level crates like `avr-hal` which implement a much easier to use interface ontop.

Rahix (commented):
Hey, thank you very much for the PR, I really like this idea and I think we should actually document a lot more than just that.  To start, I've left a few comments below.

mbuesch:
Thanks for your comments.
I implemented all your suggestions. And I also did the following changes:

* Collapsed the supported device list. Now it takes less valuable first-sight screen space. But it's a bit harder to find a specific device in the list. What do you think? I think it's worth the trade off.
* Added a small section talking about crate features.

src/interrupt.rs
```diff
@@ -3,9 +3,29 @@
...
+//! ```
+//!
+//! To access shared state, Mutex can be used:
+//!
+//! ```
+//! use avr_device::interrupt::Mutex;
+//! use core::cell::Cell;
+//!
+//! // Use Cell, if the wrapped type is Copy.
+//! // Use RefCell, if the wrapped type is not Copy or if you need a mut reference to it for other reasons.
```
Rahix, line 18: Actually, you will also need the RefCell to get an immutable reference...

```suggestion
//! // Use RefCell, if the wrapped type is not Copy or if you need a reference to it for other reasons.
```

Rahix:
> Collapsed the supported device list. Now it takes less valuable first-sight screen space. But it's a bit harder to find a specific device in the list. What do you think? I think it's worth the trade off.

Agreed, I also prefer it like this.  If people need to find a device, they can `CTRL-F`...

> Added a small section talking about crate features.

Also a good idea!  Thanks!

src/interrupt.rs
mbuesch: I think we don't need a (Ref)Cell at all, if only an immutable ref is needed. Mutex::borrow() returns an immutable ref.
In fact, we wouldn't even need a Mutex for immutable-only global access.
Or am I missing something?

Rahix: Right, but at some point you'll want to fill the Mutex so you need the mutable access at least once.  I don't think there is way around a RefCell then, even if you only grab immutable references later...

mbuesch: Ok. It could be static initialization. But this is theoretical anyway. Almost all uses would need Cell or RefCell mutable. Therefore, I removed the word.

Rahix (approved):
Perfect, thanks a lot!

PR #105, opened 2022-08-07
RFC: Add statically checked macro for safe struct Peripherals creation

mbuesch:
The `Peripherals::take()` function has two disadvantages:
* It uses one byte of precious RAM for runtime memory safety checking.
* It uses a couple of bytes of precious program memory for establishing a critical section.

This change uses the linker to statically ensure that only one Peripherals instance is constructed globally.
It introduces a new macro to avr_device that is used just like:

```
let dp = avr_device::peripherals!(atmega328p);
```

In case of multiple macro invocations, the linker will abort with

```
multiple definition of `__ERROR__avr_device__peripherals__macro_must_only_be_used_once__'
```

TODO:
* For actually not wasting the one byte of memory, the `DEVICE_PERIPHERALS` access has to be removed from the generated `Peripherals::steal()` function.
* In order to be sound, all `Peripherals::take()` function have to be removed from the generated code.


mbuesch:
As we need support from the `svd2rust` tool to properly implement this, the macro should probably also be moved into the generated code.
The `svd2rust` could get a new option to select between static and dynamic peripherals struct checking. `avr-device` could then always request generation of the static macro.

Rahix:
Hi, sorry, I completely lost track of this.  I agree with you, I think this should be discussed with the svd2rust project and integrated there.

PR #106, opened 2022-08-08
Remove all llvm_asm macros

mbuesch:
Remove all `llvm_asm!` macros and require a recent compiler with `asm!` support.
As discussed in #104


mbuesch:
Update: Removed the compile_error, because it doesn't work. The compile aborts at the earlier feature request.

mbuesch:
CI fails due to ancient compiler `nightly-2021-01-07`, as expected.

mbuesch:
Update: Remove now unnecessary build dependency `rustversion`.

src/lib.rs
```diff
@@ -131,8 +131,7 @@
 //! * To enable the crate's runtime environment, use the `rt` feature.
 
 #![no_std]
-#![cfg_attr(avr_device_asm_macro, feature(asm_experimental_arch))]
-#![cfg_attr(not(avr_device_asm_macro), feature(llvm_asm))]
+#![feature(asm_experimental_arch)] // For asm! macro.
```
Rahix, line 134: Hmm, I feel like we should only enable this when building for avr, so something like

```rust
#![cfg_attr(target_arch = "avr", feature(asm_experimental_arch))]
```

Rahix (commented):
Hey, thanks for taking care of this!  I have one minor nit, everything else looks great!

Rahix:
Ah, and CI is failing because it is still using the old nightly.  I'll push a commit to upgrade it, then you can rebase to (hopefully) get it to pass.

Rahix:
Done in commit 2e8d033ae636 ("ci: Upgrade to a new compiler version").  Please rebase when you find the time :)

mbuesch:
Thanks. I pushed an updated branch.

Rahix (approved):
Thanks!

PR #108, opened 2022-08-09
Update the bare-metal crate to 1.0.0

mbuesch:
The new crate version generates much cleaner code on AVR due to `#[inline]` annotations.

This is an API break, because the `CriticalSection` + `Mutex` mechanism changes
from exterior reference lifetime to interior reference lifetime management.

Also bare-metal seems to be working on a simplified API for `Mutex<RefCell>`. That is not included in 1.0.0, yet. But I think we should make the API breaking change now along with all other API breaking changes.

This PR is meant to be re-based after #104 is merged.


Rahix (approved):
Thanks!  I think in the wider Rust-Embedded community, [the plan is to deprecate `bare-metal`](https://github.com/rust-embedded/wg/blob/1296d1d08a7854bd9469ba2faa7e1889376d83c4/minutes/2022-08-16.md?plain=1#L59) but I think as a first step, upgrading is a good idea.

mbuesch:
Yeah, I know.
We should probably migrate to the `critical-section` crate. But upgrading `bare-metal` first is a step towards it. It makes the API a bit more `critical-section`-like and it really improves the generated code a lot.

Rahix:
Rebase when you have time, then I'll merge this ASAP :)

mbuesch:
Done.

PR #109, opened 2022-09-08
Cargo.toml: Update edition to 2021

mbuesch:
We have recently (#104) dropped support for compilers that don't support edition 2021.
This PR makes this explicit up enforcing edition 2021 in Cargo.toml.


Rahix (approved):
Perfect, thanks!

PR #110, opened 2022-09-09
Small documentation fixes

mbuesch:
Some minor fixes, that slipped through before release.

* Mutex::borrow() cs parameter is not a reference (in new bare-metal).
* Mutex example: Use proper code in a function body that would actually compile.
* Document the `ufmt` feature.
* Cargo example version

Shall we also bump the macros edition to 2021?


Rahix (approved):
Thanks!  Pulling the macros crate to 2021 is also a good idea.

mbuesch:
Thanks. I updated the PR to include the edition change for macros.

PR #112, opened 2022-12-21
Add patches for ATmega8

Rutherther:
This PR adds patches for ATmega8 needed to implement ATMega8 to avr-hal easily.

Unfortunately I had to edit the vendor file, because there is a field WGM2 consisting of WGM21 and WGM20 that are not directly next to each other. Any idea on how to patch this correctly?

### DIFF
--- a/patch/atmega8.yaml
+++ b/patch/atmega8.yaml
@@ -78,7 +78,7 @@ WDT:
     WDPL:
       CYCLES_16K:      [0, "- 16K (16384) cycles, ~16ms"]
       CYCLES_32K:      [1, "- 32K (32768) cycles, ~32ms"]
-      CYCLES_64K:      [2, "- 64K (65536) cycles, ~65s"]
+      CYCLES_64K:      [2, "- 64K (65536) cycles, ~65ms"]
       CYCLES_128K:     [3, "- 128K (131072) cycles, ~0.13s"]
       CYCLES_256K:     [4, "- 256K (262144) cycles, ~0.26s"]
       CYCLES_512K:     [5, "- 512K (524288) cycles, ~0.52s"]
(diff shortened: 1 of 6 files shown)


Rutherther:
ATmega8 uses the same register for UCSRC and UBRRH, the way to differentiate between them is that when setting UCSRC, you have to make UMSEL (MSb) one. Is there a possibility to handle that automatically? Currently the generated code contains functions in usart.rs instead of fields
```rs
    #[doc = "0x17 - USART Control and Status Register C"]
    #[inline(always)]
    pub fn ucsrc(&self) -> &crate::Reg<ucsrc::UCSRC_SPEC> {
        unsafe {
            &*(((self as *const Self) as *const u8).add(23usize)
                as *const crate::Reg<ucsrc::UCSRC_SPEC>)
        }
    }
    #[doc = "0x17 - USART Baud Rate Register Hight Byte"]
    #[inline(always)]
    pub fn ubrrh(&self) -> &crate::Reg<ubrrh::UBRRH_SPEC> {
        unsafe {
            &*(((self as *const Self) as *const u8).add(23usize)
                as *const crate::Reg<ubrrh::UBRRH_SPEC>)
        }
    }
```
But it does not seem that it sets MSB to one automatically, that's understandable as there is nothing about that in the vendor files nor in patches.

Rahix (approved):
Hi, thanks for the PR!  The changes look good to me, any reason the PR is still marked as draft?

> I am looking at the generated code and it doesn't seem that having a field on positions not directly to each others is even possible now.

Yes, these weird split fields are unfortunately not supported...  SVD can't even represent them I think.

> ATmega8 uses the same register for UCSRC and UBRRH, the way to differentiate between them is that when setting UCSRC, you have to make URSEL (MSb) one. Is there a possibility to handle that automatically?

No, I don't think there is... HAL code will need to take care of only writing to the register with the bit set correctly.

Rutherther:
Alright, I thought these problems could be solved, that was the reason it was a draft. I think it may be merged now.

PR #122, opened 2023-02-27
Fix ADC patch for ATmega128A

DaanDekoningKrekels:
Only Single Ended Inputs yet. This addition is needed for a future commit to `avr-hal` involving `atmega-hal` and the ADC.


Rahix (approved):
Thanks!

Issue #130, opened 2023-05-23
attiny_hal::pac::exint::pcmsk::W missing BitWriterRaw implementation?


mgrunwald:
Hello,

I'm only a rookie, but this seems strange for me: When I want to change the `PCMSK` register, [there is no way to access single bits](https://rahix.github.io/avr-hal/attiny_hal/pac/exint/pcmsk/struct.W.html), like there is e.g. for the `GIMSK`:

```
            self.exint.gimsk.modify(|_, w| w.pcie().set_bit()); // Good
            self.exint.pcmsk.modify(|_, w| w.bits(1)); // Not so good
```
There aren't even constants for the single bits…

I'd expect to be able to do this:
```
            self.exint.pcmsk.modify(|_, w| w.pcint0().set_bit());
```

Am I right?

Rahix:
Transferred this issue to the `avr-device` project...

It seems the register information is incorrect here.  The way to fix is this adding a YAML patch to correct it.  For example, here we have a patch which sets a field read-only:

https://github.com/Rahix/avr-device/blob/aa5b20202063a92378baf54ba7a783a2d642dcc6/patch/common/pll.yaml#L6-L10

we'd need a similar patch to set this field to `read-write` in the patch file for the MCU you are using.  Can you try sending a pull request for this?  If you need more guidance, just let me know!

mgrunwald:
@Rahix I don't think that it's a read/write issue, but that the fields are missing. Compare with `GIMSK` from `attiny85.svd`:
``` xml
        <register>
          <name>GIMSK</name>
          <description>General Interrupt Mask Register</description>
          <addressOffset>0x26</addressOffset>
          <size>0x8</size>
          <access>read-write</access>
          <fields>
            <field>
              <name>PCIE</name>
              <description>Pin Change Interrupt Enable</description>
              <bitRange>[5:5]</bitRange>
              <access>read-write</access>
            </field>
            <field>
              <name>INT0</name>
              <description>External Interrupt Request 0 Enable</description>
              <bitRange>[6:6]</bitRange>
              <access>read-write</access>
            </field>
          </fields>
        </register>
```
`PCMSK` has no fields:
``` xml
        <register>
          <name>PCMSK</name>
          <description>Pin Change Enable Mask</description>
          <addressOffset>0x0</addressOffset>
          <size>0x8</size>
          <access>read-write</access>
          <writeConstraint>
            <range>
              <minimum>0</minimum>
              <maximum>255</maximum>
[...]

PR #131, opened 2023-06-08
Patch attiny85 PCMSK register

mgrunwald:
Fix https://github.com/Rahix/avr-device/issues/130 by adding fields for pins where the pin change interrupt can be enabled. I.e. it is possible to use this

``` rust
self.exint.pcmsk.modify(|_, w| w.pcint0().set_bit());
```
instead of this:
``` rust 
self.exint.pcmsk.modify(|_, w| w.bits(1));
```

Since the two most significant bits of PCMSK on the attiny85 are reserved, it would make sense to change the write constraint to `[0, 63]`, but I failed to achieve this. the code is commented in commit https://github.com/Rahix/avr-device/commit/78b331fd41c9228ba10e3354f5c7d1482eae47e0 . Because of this, the PR is still a draft.

### DIFF
--- a/patch/attiny85.yaml
+++ b/patch/attiny85.yaml
@@ -99,6 +99,10 @@ CPU:
       PLOCK:
         access: read-only
 EXINT:
+  # This doesn't work:
+  # _modify:
+  #   PCMSK:
+  #     _write_constraint: [0, 63]
   PCMSK:
     _add:
       PCINT0:
--- a/patch/attiny85.yaml
+++ b/patch/attiny85.yaml
@@ -98,3 +98,36 @@ CPU:
     _modify:
       PLOCK:
         access: read-only
+EXINT:
+  PCMSK:
+    _add:
+      PCINT0:
+        description: Enable pin change interrupt on pin 0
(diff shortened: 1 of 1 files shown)


patch/attiny85.yaml
```diff
@@ -98,3 +98,40 @@ CPU:
     _modify:
       PLOCK:
         access: read-only
+EXINT:
+  # This doesn't work:
+  # _modify:
+  #   PCMSK:
+  #     _write_constraint: [0, 63]
```
Rahix, line 105: Yeah, the version of `svdtools` which we are using at the moment doesn't seem to support updating the write constraint of whole registers :/

Let's keep it the way you wrote it so we can uncomment the block once we upgrade to a version of `svdtools` that does support it.

Rahix (approved):
Hi,

I'm so sorry for loosing track of this :(  Thanks a lot again for your work here!  It's looking great already.  The write-constraint thing is something I'm afraid we'll have to keep as it is for now.  This will be fixed as soon as we upgrade to a newer `svdtools` I believe.

PR #133, opened 2023-07-05
Cleanup

SnakeOilSalesman:
Hey!

I have few minor suggestions at your trial)

1. My IDE (PyCharm) complains about docstring example in entry point's docstring. Technically, since `main` function annotated with `!`, linter expects an infinite loop in it. That's the change.
2. Code-style improvements:
    1. Line length complains. Few lines are longer than 100 chars.
    2. There is a redundant blank line in one place, according to `rustfmt`.
3. During my work with this crate, few times I forgot that `svdtools` might deserve a special handling. Without it `Makefile` commands wouldn't work. Tracing back this issue is time consuming. I'd suggest to add an additional step to `README.md` to check that `svdtools` actually installed correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,13 @@ cargo install svd2rust --version 0.28.0
 cargo install form
 rustup component add --toolchain nightly rustfmt
 pip3 install --user svdtools
+
+# check svdtools
+svd --version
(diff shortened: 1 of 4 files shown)


Rahix (commented):
Very nice, thanks a lot!

PR #134, opened 2023-07-06
ATtiny828: Fix ADC reference voltage variant name

SnakeOilSalesman:
Hey!

Am currently working on a PR to add ATtiny828 to `avr-hal`. Just found out that I previously used a wrong name for an ADC ref voltage variant.

My bad, sorry for the inconvenience.


Rahix (approved):
Nothing to be sorry about, thanks for the PR!

PR #146, opened 2024-01-04
example: add simple atmega328+interrupts example

Frankkkkk:
Hi,

This commits adds a simple example to this crate. Maybe it can be useful to others; what do you think ?

Thanks and have a nice day !


examples/atmega328p/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+name = "mega328-test"
+version = "0.1.0"
+authors = ["Frank Villaro-Dixon <frank@villaro-dixon.eu>"]
+edition = "2021"
+license = "Public domain"
```
Rahix, line 6: Please use `MIT OR Apache-2.0` licensing.

examples/atmega328p/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+        });
+    }
+}
+
+#[avr_device::entry]
+fn main() -> ! {
+    let dp = avr_device::atmega328p::Peripherals::take().unwrap();
+
+    // The following should be a NO-op
+    dp.TC0.tccr0a.write(|w| w.wgm0().bits(0));
```
Rahix, line 57: The comment should instead focus on the intent of the line below.  Why are you setting the WGM to 0?

examples/atmega328p/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    // Divide by 1024 -> 16MHz/1024 = 15.6kHz
+    dp.TC0.tccr0b.write(|w| w.cs0().prescale_1024());
+    // Enable overflow interrupts
+    dp.TC0.timsk0.write(|w| w.toie0().set_bit());
+
+    // Make pd2 and pd3 outputs
+    // We use modify in order not to change the other bits
```
Rahix, line 65:
```suggestion
    // We use .modify() in order not to change the other bits
```

examples/atmega328p/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+fn main() -> ! {
+    let dp = avr_device::atmega328p::Peripherals::take().unwrap();
+
+    // The following should be a NO-op
+    dp.TC0.tccr0a.write(|w| w.wgm0().bits(0));
+
+    // Divide by 1024 -> 16MHz/1024 = 15.6kHz
+    dp.TC0.tccr0b.write(|w| w.cs0().prescale_1024());
+    // Enable overflow interrupts
+    dp.TC0.timsk0.write(|w| w.toie0().set_bit());
```
Rahix, line 62: As you're explicitly explaining `.modify()` below, I think it makes sense to add a comment somewhere up here explaining `.write()` as well.  Namely, mention that `.write()` writes the whole register to its reset value + the modified bits.

examples/atmega328p/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    // Make pd2 and pd3 outputs
+    // We use modify in order not to change the other bits
+    dp.PORTD.ddrd.modify(|_, w| w.pd2().set_bit());
+    dp.PORTD.ddrd.modify(|_, w| w.pd3().set_bit());
+
+    unsafe {
+        avr_device::interrupt::enable();
+    }
```
Rahix, line 71: `// SAFETY:` comment explaining that we are allowed to enable interrupts here because we know we are not in a critical section.

examples/atmega328p/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+        match led_state {
+            true => dp.PORTD.portd.modify(|_, w| w.pd2().set_bit()),
+            false => dp.PORTD.portd.modify(|_, w| w.pd2().clear_bit()),
+        }
+
+        // We want to make the program crash after 9 blinks
+        if previous_state != led_state {
+            counter -= 1;
+            let _ = 1 / counter; // Make it fail after 9 attempts !
+        }
```
Rahix, line 90: This is funny to people who get the division by zero but for an example I think we should make this more boring and more immediately obvious by instead doing something like

```rust
if previous_state != led_state {
    counter += 1;
}
if counter > 9 {
    // intentionally panic to show off the panic handler
    panic!();
}

File: examples/atmega328p/.cargo/config.toml [hunk 7]
```diff
@@ -0,0 +1,8 @@
+[build]
+target = "avr-specs/avr-atmega328p.json"
+
+[target.'cfg(target_arch = "avr")']
+runner = "ravedude nano -cb 57600"
```
Rahix, line 5: Please default to `uno` instead of `nano`

Rahix (changes requested):
Hi, thanks a lot for your contribution!  I like the idea of adding an example to `avr-device` which focuses purely on this crate and is unrelated to the layer above (`avr-hal`).

However, I think adding such an example isn't entirely uncontroversial:  People will stumble over this example when searching for ways to write Rust code for AVR so we need to make sure they are pointed in the right direction.

My suggestion is to add a README file to `examples/atmega328p/` which explains

1. What this example demonstrates (A program that only uses bare register writes, without a HAL abstraction)
2. Where the HAL can be found and that the HAL also has examples
3. How this example can be tried out (UNO board, ravedude, LEDs on what pins)

examples/atmega328p/Cargo.toml
Frankkkkk: If that's okay, for this small and easy example I prefer the public license as IMHO it makes it less cumbersome for everyone. Of course this doesn't change the license of the library.

Frankkkkk:
Hi @Rahix Thanks for your great feedback. I've updated the example. Happy to get your feedback.
Cheers !

examples/atmega328p/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+
+
+[dependencies]
+ufmt = "0.2.0"
+nb = "0.1.2"
+embedded-hal = "0.2.3"
+
+
+[dependencies.avr-device]
+version = "0.5.3"
```
Rahix, line 21: Sorry, didn't catch this the first time around:

Including `avr-device` from crates.io is a bit unusual as you'd usually want the examples of a library crate to use the crate that's part of the same repository.  This is very useful if you want to try an example against changes in the library code.

You could do this by replacing the `version = ` with a `path = "../.."` line.

**However**, for `avr-device` in particular, this isn't quite that simple:  The problem is that to use the repository version of `avr-device`, you first have to run all the `make` steps to generate the crate's code.  I wouldn't want that to be required for just running an example...

So my suggestion is this:  Let's keep the `version = ` line but add the `path = ` line next to it, in a comment.

```suggestion
version = "0.5.3"
# To use the local version of avr-device instead, uncomment the following line:
path = "../.."
```

examples/atmega328p/Cargo.toml
Rahix: I totally get your intention here (and would love to accept it this way), but I am asking for this change for practical reasons:  Introducing a second license here means that diffent parts of the repository are licensed differently and that's always a headache.  It means, for example, you couldn't simply copy code from an example in the crate's documentation into the example.

Also, public domain licensing has very unclear legal status in a lot of jurisdictions which further complicates things.

You can still license your contribution under any alternative terms everywhere else, but for inclusion in this repository, it'll make my life a lot easier to stick to the crate license :)

examples/atmega328p/README.md
```diff
@@ -0,0 +1,82 @@
...
+The main loop just watches the state of the `LED_STATE` variable and sets the green led (PD3) 
+according to the state of the boolean.
+
+Every time this variable changes, a counter is incremented. When it reaches 10, a panic is raised
+and our custom panic handler is called.
+
+
+## Trying it out
+
+Change into the example directory:
```
Rahix, line 63: I think we should explain what dependencies need to be installed to build the example.  The easiest way to do this is probably to link this section from the `avr-hal` README: https://github.com/Rahix/avr-hal#quickstart

examples/atmega328p/src/main.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        // We want to make the program crash after 9 blinks
+        if previous_state != led_state {
+            counter += 1;
+        }
+        if counter > 9 {
+            // The following simulates a panic, but it could also be
+            // a more "general" bug like dividing by zero, out of bounds,
+            // etc..
+            panic!();
```
Rahix, line 98: Nit-pick:  This doesn't simulate a panic, it really _is_ a panic.  But you don't need to change this unless you really want to...

examples/atmega328p/src/main.rs
```diff
@@ -0,0 +1,102 @@
...
+    loop {
+        let mut led_state: bool = true;
+        interrupt::free(|cs| {
+            led_state = LED_STATE.borrow(cs).get();
+        });
+
+        match led_state {
+            true => dp.PORTD.portd.modify(|_, w| w.pd2().set_bit()),
+            false => dp.PORTD.portd.modify(|_, w| w.pd2().clear_bit()),
+        }
```
Rahix, line 88: Now that I'm reading it again, I just noticed it:  You can simplify this a bit by doing the following instead:

```suggestion
        dp.PORTD.portd.modify(|_, w| w.pd2().bit(led_state));
```

Frankkkkk:
Thanks for the review !

examples/atmega328p/README.md
```diff
@@ -0,0 +1,85 @@
...
+The main loop just watches the state of the `LED_STATE` variable and sets the green led (PD3) 
+according to the state of the boolean.
+
+Every time this variable changes, a counter is incremented. When it reaches 10, a panic is raised
+and our custom panic handler is called.
+
+
+## Trying it out
+
+First of all, read the readme of the [avr-device](https://github.com/Rahix/avr-hal#quickstart)
```
Rahix, line 63: It should be `avr-hal`, not `avr-device` here ;)

```suggestion
First of all, read the README of the [avr-hal](https://github.com/Rahix/avr-hal#quickstart)
```

Rahix (commented):
Thanks for the changes!  I have one more tiny nit, then we're good to merge this.  Please check my comment below.

Frankkkkk:
done :)

Rahix (approved):
Thanks a lot!

PR #152, opened 2024-04-10
Add support for new AVR USB devices. AVR64DU32 and AVR64DU28.

jwagen:
Adds basic support for AVR64DU32 and AVR64DU28. I have only spot checked the generated doc, they look ok.

https://www.microchip.com/en-us/product/AVR64DU32


Rahix (approved):
Thanks!

PR #156, opened 2024-05-04
ci: Bump nightly toolchain

Rahix:
Use the same version that is now used by avr-hal.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,12 +66,12 @@ jobs:
       - name: Install Nightly Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          toolchain: nightly-2023-08-08
+          toolchain: nightly-2024-03-22
           components: rustfmt
 
       # Actual test run
       - name: Generate chip description sources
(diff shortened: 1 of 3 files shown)


Rahix:
Hm, so we are running into https://github.com/Rahix/avr-hal/issues/537 here as well now.  I'm considering whether going "nuclear" and just pinning `proc-macro2` for `avr-device-macros` directly might be the best solution:

```diff
diff --git a/macros/Cargo.toml b/macros/Cargo.toml
index 617e80b..5c3ece9 100644
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -17,7 +17,7 @@ proc-macro = true
 
 [dependencies]
 quote = "1.0.7"
-proc-macro2 = "1.0.19"
+proc-macro2 = "=1.0.79"
 
 [dependencies.syn]
 version = "1.0.35"
 ```

 Please voice your opinions...

Rahix:
A downside of pinning in `avr-device-macros` is that downstream users cannot easily override it, I think.  This means that when people want to switch to a different toolchain version, they can't adjust the `proc-macro2` pin to work alongside it... Unfortunate :(

PR #163, opened 2024-10-16
Added basic mods for atmega16u2

zacharytomlinson:
make atmega16u2
-e      SVD2RUST        atmega16u2
error: a value is required for '-i <FILE>' but none was supplied


Rahix:
Hi,

I think the problem is that your ATDF is called `atmega16U2` instead of `atmega16u2`...

Rahix:
I pushed commit 5a586731107c ("make: Fix make continuing when the ATDF source file is missing") to have `make` error out instead of `svd2rust` when this happens.  But the error message is still not great due to my make magic not being good enough...

zacharytomlinson:
Jesus your right.... That's the stupidest mistake I've made in at least the last 12 minutes lol. I changed the filename and now it is building. How would you prefer for me to clean up? Mark the issue complete, close this draft PR and open a new one, or just update this PR?

Rahix:
Well, the error message could have been way more clear - there's nothing hinting at a file not being found...  So don't be too hard on yourself ^^

---

Usually, it would have been just fine to continue with this PR, and then mark it as ready for review once you are done.  But you can also open a new one, don't worry :)

zacharytomlinson:
I am going to do a once-over of the patch to make sure the basics are all in there and there are no glaring issues and then I will open a new one!

PR #164, opened 2024-10-17
Add ATmega32U2, ATmega16U2, ATmega3208, and ATmega3209

zacharytomlinson:
Added deps for 32u2 target, fixed accidental import of 16u2 target when targeting atmega64 and added note to readme to check downloaded atdf file names as the capitalization is sometimes wonky.


zacharytomlinson:
NOTES: 
- I added a note in the readme to check downloaded atdf file names as they are capitalized weird in both sources. 
- Do we still need the impl for Peripherals in the devices/mod.rs? The change log for the atmega64 updates has that in 
there but no other chip implementations have those anymore.

zacharytomlinson:
Is there anything else you need me to do here @Rahix?

Rahix (approved):
Hi,

I've taken the liberty to clean up your commits such that each commit contains a logical change.  In the future, please don't put so many things into a single PR, instead send multiple PRs — one for each chip, for example.

The overall changes themselves look fine, thanks a lot for your contribution!

PR #166, opened 2024-12-01
Add support for ATmega88p

blueberry-fan:
This add support for ATmeg88p.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,6 +82,6 @@ ufmt = { version = "0.2.0", optional = true }
 critical-section = { version = "1.1.1", optional = true }
 
 [dependencies.avr-device-macros]
-path = "macros/"
+path = "macros"
 version = "=0.6.0"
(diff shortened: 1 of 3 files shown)


Rahix (approved):
Ah, interesting, so these changes were not actually added in #120 so ATmega88P was not actually usable so far.

Thanks for fixing this!

blueberry-fan:
When do you plan to issue a new version of avr-device? 
I ask because of https://github.com/Rahix/avr-hal/pull/608, which depends on this.

tones111:
Unrelated I just pushed a review to bump `avr-hal` to [use version 0.6](https://github.com/Rahix/avr-hal/pull/618).  If the new `avr-device` gets a minor version update (to 0.6.1) then that MR still has value.  Otherwise I can update it to use 0.7.  If that's the case then lets add a comment there to wait for the new `avr-device` to get published.

PR #169, opened 2025-01-04
Refactored attiny-xmega patch files

innermatrix:
A light refactor of the patch files. Since the ATtiny xmega series have a lot in common, it makes more sense to structure their patch files as "common + 0 series" or "common + 1 series" than "0 series" or "0 series + 1 series"


Rahix (approved):
Thanks a lot!

PR #171, opened 2025-01-04
Rename in to input in ATtiny xmega PORT

innermatrix:


Rahix (approved):
Fair enough, makes sense.  Do keep in mind that we should stay as close as possible to the "upstream" register/field naming so cross-referencing with datasheets is always easily possible.  I do agree that this is a sensible change though.

innermatrix:
Yeah this was the smallest thing that I could think of that didn't require quoting and wouldn't be too surprising to humans.

PR #173, opened 2025-01-05
Update some peripherals for Atmega128rfa1

NikBel3476:
In `timer/atmega128rfa1.yaml` "_replace_enum" from `timer/dev/*.yaml` applies only for first COM* definition (svdtools 0.1.27) so I added it manually. I don't know what the reason but patch works correctly without manual definitions with svdtools 0.3.20 installed from cargo.

### DIFF
--- a/patch/atmega128rfa1.yaml
+++ b/patch/atmega128rfa1.yaml
@@ -16,3 +16,748 @@ _include:
   - "common/twi.yaml"
   - "common/usart.yaml"
   - "common/wdt.yaml"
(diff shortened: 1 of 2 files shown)


tones111:
Maybe we can update to the rust version of svdtools (#174)?

NikBel3476:
I just checked it out and it doesn't depend on changes in #174.

Rahix (approved):
Alright, thanks a lot for your contribution!

PR #174, opened 2025-01-05
svdtools: install rust version

tones111:
This change removes a dependency on python/pip and is more consistent with the installation of the other supporting tools.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,21 +20,13 @@ jobs:
         with:
           toolchain: stable
 
-      # Python Dependencies
-      - name: Install Python dependencies
-        run: |
-          pip3 install --user setuptools wheel
-          pip3 install --user svdtools
-      - name: Put pip binary directory into path
-        run: echo "$HOME/.local/bin" >> $GITHUB_PATH
-
       # Rust Dependencies
       - name: Cache Cargo installed binaries
         uses: actions/cache@v4
         id: cache-cargo
         with:
           path: ~/cargo-bin
-          key: rust-tools-20250104-001
+          key: rust-tools-20250105-001
       - name: Install svd2rust
         if: steps.cache-cargo.outputs.cache-hit != 'true'
         run: cargo install svd2rust --version 0.28.0 --locked
@@ -44,6 +36,9 @@ jobs:
       - name: Install atdf2svd
         if: steps.cache-cargo.outputs.cache-hit != 'true'
         run: cargo install atdf2svd --version 0.5.0 --locked
+      - name: Install svdtools
+        if: steps.cache-cargo.outputs.cache-hit != 'true'
+        run: cargo install svdtools --version 0.3.21 --locked
       - name: Copy tools to cache directory
         if: steps.cache-cargo.outputs.cache-hit != 'true'
         run: |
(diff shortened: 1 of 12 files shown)


tones111:
The CI needs the cache cleared but I'm not familiar with how to do that.

Rahix:
You have to bump this cache key to have CI generate a new tools cache:

https://github.com/Rahix/avr-device/blob/07f32b878717ea14922f47ba4db03048436d20dd/.github/workflows/ci.yml#L37-L37

Rahix:
You can download the CI artifacts from main branch, e.g. here: https://github.com/Rahix/avr-device/actions/runs/12615563082/artifacts/2386623070

Rahix:
Do post the the diff of the generated svds in this pull request please :)

Rahix:
@tones111, check the changes in this PR https://github.com/Rahix/avr-device/pull/157/files, I think a large part of the patch file porting has already been done there.

tones111:
Took a bit to figure out the yaml syntax and why it was complaining but the changes to resolve ended up being pretty straightforward.  I committed the generated svd files, so once we've looked at them I will remove it.

When comparing against the svd files @Rahix linked above there were some formatting improvements.  Running the original patched files through `xmllint --format` really helped.  Doing some spot checks it looks like the main difference is that the new `svdtools` removes the `<usage>...</usage>`  from enumerated values (which makes sense).

tones111:
[avr-device.svd.diff.txt](https://github.com/user-attachments/files/18319682/avr-device.svd.diff.txt)

Here's the diff against an `xmllint` version of 0.7.0.  I see the following differences:

- `<usage>...</usage>` removed from enumerals
- improved description whitespace
- unicode improvements (ex: atmega128rfa1)
- Some ISC1 interrupt enumerals are removed (ex: attiny167, attiny2313, etc)
- Some ADC enumerals changed in attiny84

Rahix:
> Some ISC1 interrupt enumerals are removed

They are removed because the derived enumerals from ISC0 already include properly named values for them.  This is still a breaking change, but one we do want in my eyes.

> Some ADC enumerals are reordered in attiny84

Looks okay to me.

---

As a final check, you can diff the generated rust sources of each version.  You can find the original ones in the crate archive contained in `target/package/` dir of the artifacts archive.  The `.crate` file is just a `.tar.gz` archive of the sources.

I would expect this diff to only show the whitespace changes and the removed ISC1 enumerated values.

Rahix (approved):
Changes look fine, after the aforementioned cross-check of the generated sources and removal of the comitted SVDs, this is ready for merge :)

BTW, comitting the SVDs wasn't necessary - CI again created an artifacts archive that includes them for your PR: https://github.com/Rahix/avr-device/actions/runs/12625939793?pr=174

Rahix:
Ah, just noticed [svdtools version 0.4.0](https://github.com/rust-embedded/svdtools/releases/tag/v0.4.0) is hot off the stove, maybe we can update this PR to the new version immediately to reduce work later on? :)

tones111:
Here's a new diff (against main svds) for the update to svdtools 0.4.0.
[svdtools_040.diff.txt](https://github.com/user-attachments/files/18325747/svdtools_040.diff.txt)

A few more tweaks were required that resulted in some minor corrections
- updated atmega328pb TWAM1 description
- atmega8 UPE field is named PE (matches datasheet)
- atmega841 ACO{0,1} are read-only (matches datasheet)

However, it looks like the timer enum changes from #173 aren't working but I'm not sure why.

tones111:
Maybe the atmega128rfa1 enum changes are there.  The svd file has
```
<enumeratedValues derivedFrom="COM0B"/>
```

and the `COM0B`  looks like it matches the patch file.  So, I think the new version of svdtools knows how to deduplicate the enums.

Rahix (approved):
Thanks so much for taking care of this!

PR #155, opened 2024-05-04
Update to svd2rust 0.33.1

Rahix:
This improves the generated API in various ways.

The most major change seems to be from [0.31.0](https://github.com/rust-embedded/svd2rust/releases/tag/v0.31.0) where registers are now accessed using methods instead of fields.  This will require rather large downstream changes.

Upstream svd2rust also switched to a different case style for identifiers in [0.32.0](https://github.com/rust-embedded/svd2rust/releases/tag/v0.32.0).  For now, this is disabled using `--ident-formats legacy` but we should consider following their suggestion at some point in the future.

PR #176, opened 2025-01-11
svd2rust: update to v0.35.0

tones111:
This is an effort to bridge the gap between #155 and #157.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,10 +26,10 @@ jobs:
         id: cache-cargo
         with:
           path: ~/cargo-bin
-          key: rust-tools-20250106-001
+          key: rust-tools-20250111-001
       - name: Install svd2rust
         if: steps.cache-cargo.outputs.cache-hit != 'true'
-        run: cargo install svd2rust --version 0.28.0 --locked
+        run: cargo install svd2rust --version 0.35.0 --locked
       - name: Install cargo-form
         if: steps.cache-cargo.outputs.cache-hit != 'true'
         run: cargo install form --version 0.8.0 --locked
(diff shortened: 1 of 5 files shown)


tones111:
hmm.  The example built ok locally after updating the dependency to use the path.  How do I teach the CI how to do that?

stappersg:
afbeelding

> hmm. The example built ok locally after updating the dependency to use the path. How do I teach the CI how to do that?
> 
> Or maybe the trick is to commit the change with the line un-commented, and then re-comment it on a followup commit.

Other idea:  Just _"svd2rust: update to v0.35.0"_ in this merge request and the ATmega328P changes in later merge request.

Plus accepting that _github CI_ is an aid and not a goal.  I mean: We want to move forward, even stumbling forward. _github CI_ should never block that.

LuigiPiucco:
CI is failing because the published version of avr-device still has the uppercase identifiers, but you lowercased them in the example (it is a separate crate before #157, so it doesn't sync with your local changes unless you use a path dependency). I think you'll have to update the example only after publishing a new version to avoid the path dependency, if you want that.

.github/workflows/ci.yml
```diff
@@ -86,10 +86,43 @@ jobs:
       - name: Install Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          toolchain: nightly-2023-12-28
+          toolchain: nightly-2025-01-10
           components: rust-src,rustfmt
+
+      # Rust Dependencies
```
tones111, line 92: The duplication here feels pretty dirty, but maybe there's a better way for the jobs to depend on a separate job that builds the cache.

Rahix:
For reference, I'm still in preference to holding off on this if at all possible, see <https://github.com/Rahix/avr-device/pull/157#discussion_r1912489899>.

If we do have to go ahead with the update, we'll need to do another change analysis/documentation like in #155.

tones111:
I don't understand why we would lump this into #157 if we can break it out separately.  By working this here we can handle impacts related to the version change in isolation from the complexity of moving toward the build.rs solution. I'm concerned the level of complexity is too high there (see my comments in that thread about the atdf2svd version regression, missing SPI_MASTER enumeral, and commits not building).

I think it's preferable to land this more concentrated effort first since it needs to be done anyway.  If we're taking `svd2rust` as a dependency then it needs to be kept updated.  This improves consistency with the other PACs using this tooling and forking would expend a lot of additional resources.

Landing this upstream makes verification of #157 easier because the resulting diff of the output becomes much smaller.

tones111:
overcome by #157

PR #178, opened 2025-01-12
remove form dependency

tones111:
This is another stepping stone toward #157. I did not see any difference in build time on my machine when running `cargo check --all-features`.

It probably needs the same CI changes to build the example as #176, but I didn't want to duplicate them here.  This change can be rebased once that one lands.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,13 +26,10 @@ jobs:
         id: cache-cargo
         with:
           path: ~/cargo-bin
-          key: rust-tools-20250106-001
+          key: rust-tools-20250112-001
       - name: Install svd2rust
         if: steps.cache-cargo.outputs.cache-hit != 'true'
         run: cargo install svd2rust --version 0.28.0 --locked
-      - name: Install cargo-form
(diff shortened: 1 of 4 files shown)


Rahix:
Hi,

I am not sure I understand the intention of this PR.  #157 will drop `form` anyway because it removes all things referencing it.  So this PR would just generate merge conflicts for #157?

Also, removing form without a replacement does not make sense to me either - as long as #157 is not yet merged, we _want_ to have `form` to cleanup up the file structure, no?

tones111:
By landing this change upstream we reduce the complexity and diff that needs to be analyzed when evaluating #157 which has already shown to be too complicated.  I think it should focus on the Makefile -> build.rs migration as we can iron out these pieces separately.

I'm not sure I understand the motivation to run `form` against the generated mcu code.  Several years ago the compiler and `rustfmt` would choke (take several minutes) on the large modules out of svd2rust, but today the compile time is the same.  So I don't know what benefits we get from splitting the file into many modules.  I'm assuming the generated documentation is the same either way.

tones111:
#157 is getting close, so defer to that solution

PR #182, opened 2025-03-15
Rounding out ATtiny 14-pin 0-series with 204/804/1604 support

nonik0:
Adding support for the rest of ATtiny 14 pin 0-series , ATtiny202, ATtiny804, and Attiny1604. Verified basic functionality with attiny1614 with no issues so far. I was halfway working through this PR when noticed #181, so would like to acknowledge @Phil-Hacker's work as well.

I have support in avr-hal in progress as well: https://github.com/nonik0/avr-hal/tree/attiny204-804-1604

Need a bit a clarification on the guidance around maintaining the alphabetical sorting of the chips. I was getting confused and just asked an LLM to do it and it ended up reordering a lot more than I expected. So it seems like the ordering is inconsistent in various locations.


Rahix:
Hi, thanks a lot for your contribution and sorry for the late reply! Just a heads-up, I am holding off merging new chips right now until #157 is merged, which quite significantly changes the way the register definitions are generated. I will ping you once this change has landed so you can rebase your changes ontop of it.

---

> Need a bit a clarification on the guidance around maintaining the alphabetical sorting of the chips.

Yeah, there is no logic to it, unfortunately many people didn't place their chips in the right section which I then didn't want to fix to not cause unnecessary rebase work for others...  But I think the aforementioned #157 cleans up that mess to some degree as well.

Rahix:
Here's the notification: #157 was merged :tada: Please rebase your work ontop of the new code-generation.  You should be able to make out the new structure in `src/devices/mod.rs` from the way things look for the existing chips.  The `Makefile` is dropped completely.

Rahix (approved):
Thanks a lot!

PR #188, opened 2025-04-20
build: only parse selected MCU inputs

### LINKED ISSUE #187: Build script should only parse ATDF for selected MCUs
@LuigiPiucco, so I just noticed one more thing that slipped through review ;)

In `find()`, you already parse **all** ATDFs (and also all patches).  This happens before filtering for selected MCUs which means that a lot of unnecessary parsing work is done for MCUs that won't actually get code-generated.

I noticed this through suspiciously long build-times for single-MCU builds and also while playing with the error messages (injecting an error into the ATDF of an MCU that is not selected still leads to to a build error).

Do you want to take care of this?

LuigiPiucco:
Fix #187.

Since I had some extra time, I went ahead and took care of this. It's a simple change, I merged `InputFinder::filter_from_features` and `InputFinder::find`. We still iterate every path, to add the name to the list of available MCUs in case of error, but the inputs are only inserted if the variable is present.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -16,9 +16,6 @@ fn main() -> Result<(), anyhow::Error> {
     let crate_root = PathBuf::from(env::var_os("CARGO_MANIFEST_DIR").ok_or(anyhow::anyhow!(
         "env variable CARGO_MANIFEST_DIR is missing"
     ))?);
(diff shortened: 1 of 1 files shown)


Rahix (approved):
Very nice, thanks a lot!

PR #189, opened 2025-04-23
asm: Add function to get stack size

### LINKED ISSUE #175: Add function to get stack size
We should provide a function that returns the current stack size for debug analysis purposes.

(From <https://github.com/Rahix/avr-hal/discussions/589>)

LuigiPiucco:
Fix #175.

The check with `u8::MAX` is there because some MCUs (the really small ones whose data space address fits in a `u8`) may not have `SPH`. Because `extern static`s are not as optimizable as `const`s, the resulting binary can't eliminate the opposing branch, something that could be worked on.

### DIFF
--- a/src/asm.rs
+++ b/src/asm.rs
@@ -39,6 +39,53 @@ pub fn wdr() {
     }
 }
 
+/// Get the stack size in bytes, for debbuging.
+#[cfg(feature = "rt")]
+#[inline(always)]
+pub fn get_stack_size() -> usize {
+    extern "C" {
+        static __DATA_REGION_ORIGIN__: usize;
+        static __DATA_REGION_LENGTH__: usize;
+    }
+    use core::ptr::addr_of;
+
+    cfg_if::cfg_if! {
+        if #[cfg(target_arch = "avr")] {
+            unsafe {
+                let data_region_end = addr_of!(__DATA_REGION_ORIGIN__) as usize + addr_of!(__DATA_REGION_LENGTH__) as usize;
+                let sp: usize;
+                if data_region_end > u8::MAX as usize {
+                    // Here the stack pointer is 2 bytes.
+
+                    let spl: u8;
+                    let sph: u8;
+                    core::arch::asm!(
+                        "in {}, 0x3d",
+                        "in {}, 0x3e",
+                        out(reg) spl,
+                        out(reg) sph,
+                        options(nostack, nomem, pure),
+                    );
+                    sp = usize::from_le_bytes([spl, sph]);
+                } else {
+                    // Here the stack pointer is 1 byte.
+
+                    let spl: u8;
+                    core::arch::asm!(
+                        "in {}, 0x3d",
+                        out(reg) spl,
+                        options(nostack, nomem, pure),
+                    );
+                    sp = spl as usize;
+                }
+                data_region_end - sp
+            }
+        } else {
+            unimplemented!()
+        }
+    }
+}
+
 /// Blocks the program for at least `cycles` CPU cycles.
 ///
 /// This is intended for very simple delays in low-level drivers, but it


src/asm.rs
```diff
@@ -39,6 +39,51 @@ pub fn wdr() {
...
+pub fn get_stack_size() -> usize {
+    extern "C" {
+        static __DATA_REGION_ORIGIN__: usize;
+        static __DATA_REGION_LENGTH__: usize;
+    }
+
+    cfg_if::cfg_if! {
+        if #[cfg(target_arch = "avr")] {
+            unsafe {
+                let data_region_end = __DATA_REGION_ORIGIN__ + __DATA_REGION_LENGTH__;
```
Rahix, line 53: Does this actually work?  I would expect that you need to take a reference/pointer of these symbols and cast its address to a usize...

LuigiPiucco: I haven't checked on hardware, but it seems to link properly.

src/asm.rs
```diff
@@ -39,6 +39,51 @@ pub fn wdr() {
...
+                    let spl: u8;
+                    let sph: u8;
+                    core::arch::asm!(
+                        "in {}, 0x3d",
+                        "in {}, 0x3e",
+                        out(reg) spl,
+                        out(reg) sph,
+                        options(nostack, nomem, pure),
+                    );
+                    sp = usize::from_le_bytes([spl, sph]);
```
ckoehne, line 67: A trick to get the stack pointer without asm at all is reading the address from a stack variable (I've used inline never to get a better asm dump):
```
#[inline(never)]
fn get_stack_ptr() -> usize {
    let stack_val: u8 = 0;
    &stack_val as *const u8 as usize
}
```
Unfortunately, this generates quite inefficient code:
```
  6e:   cf 93           push    r28
  70:   df 93           push    r29
  72:   cd b7           in      r28, 0x3d       ; 61
  74:   de b7           in      r29, 0x3e       ; 62
  76:   21 97           sbiw    r28, 0x01       ; 1
  78:   0f b6           in      r0, 0x3f        ; 63
  7a:   f8 94           cli
  7c:   de bf           out     0x3e, r29       ; 62
  7e:   0f be           out     0x3f, r0        ; 63
  80:   cd bf           out     0x3d, r28       ; 61
  82:   ce 01           movw    r24, r28
  84:   01 96           adiw    r24, 0x01       ; 1
  86:   21 96           adiw    r28, 0x01       ; 1
  88:   0f b6           in      r0, 0x3f        ; 63
  8a:   f8 94           cli
  8c:   de bf           out     0x3e, r29       ; 62
  8e:   0f be           out     0x3f, r0        ; 63
  90:   cd bf           out     0x3d, r28       ; 61
  92:   df 91           pop     r29
  94:   cf 91           pop     r28
  96:   08 95           ret
```

src/asm.rs
```diff
@@ -39,6 +39,51 @@ pub fn wdr() {
...
+
+                    let spl: u8;
+                    core::arch::asm!(
+                        "in {}, 0x3d",
+                        out(reg) spl,
+                        options(nostack, nomem, pure),
+                    );
+                    sp = spl as usize;
+                }
+                data_region_end - sp
```
ckoehne, line 79: Doesn't the stack grow from top and the data from bottom? Shouldn't it be `sp - data_region_end`?

src/asm.rs
Rahix: I think I would stick to the inline assembly.  Is there any compelling reason not to?

src/asm.rs
Rahix: It links, but probably won't work.  What happens here is that you now read the (`usize`) values _at_ the `__DATA_REGION_ORIGIN__`, so the first bytes of SRAM and at address `__DATA_REGION_LENGTH__` which in this case will most likely be out of bounds and raise a CPU fault.

These two symbols are sort of "fake".  They are placed by the linker at specific addresses so you can use those addresses in your code.  You are never meant to "dereference" these symbols and actually read from the contained addresses.  So to use them correctly, you have to take references to those symbols, which the linker will then relocate to the respective address value.

For reference, check out this piece of code from `cortex-m-rt`: https://github.com/rust-embedded/cortex-m/blob/c3d664bba1148cc2d0f963ebeb788aa347ba81f7/cortex-m-rt/src/lib.rs#L1096-L1109

src/asm.rs
ckoehne: There are the common reasons to avoid inline assembly. It's less readable, so harder to maintain and it's highly architecture specific. My suggestion would make the code shorter, avoids any if's for sp register size and will even work for other architectures not only avr. It's most probably slower because it looks like the compiler isn't able to optimize it well. Given the issue that this commit should solve, it's only for debugging purposes, so speed shouldn't matter here.

It's just a suggestion, so go with the implementation you like! I would propose following solution:
```
fn get_stack_size() -> usize {
    let stack_var: u8 = 0;
    let sp = &stack_var as *const u8 as usize;
    let data_region_end = ...;

    data_region_end - sp
}
```

src/asm.rs
ckoehne: Sry, didn't get that data_region_end is supposed to contain ram_end.

src/asm.rs
ckoehne: You could even omit the dummy variable by using the address of data_region_end.

src/asm.rs
LuigiPiucco: I've updated the code to use `core::ptr::addr_of`.

src/asm.rs
Rahix: Honestly, I don't see how this is much better.  You are relying on some unspoken guarantee that `stack_var` will really end up on the stack, as the bottom-most byte.  But who says this function isn't inlined and `stack_var` reordered into some alignment niche?  Or spilled into rodata because it is a known read-only reference to a null-byte?  rustc/LLVM don't make any promises about this at all - for all they know, there isn't even a stack as such in the memory model.

The inline assembly is going to deterministically yield the right value, no optimizer tricks can break that.  Also, I don't really see a maintenance hazard with this tiny piece of inline assembly to read out one single register.  We have way worse inline-asm for cycle-accurate delays already...

src/asm.rs
```diff
@@ -39,6 +39,52 @@ pub fn wdr() {
     }
 }
 
+/// Get the stack size in bytes, for debbuging.
+#[inline(always)]
+pub fn get_stack_size() -> usize {
```
Rahix, line 44: Please mark `get_stack_size()` with `#[cfg(feature = "rt")]`.  This is necessary because we must not make assumptions about the "runtime" (which includes the linker script and its symbols in this case) when `rt` is not selected.

Rahix (changes requested):

src/asm.rs
LuigiPiucco: Good catch. It's done.

Rahix:
I was about to try testing this on real hardware, then I noticed we first have to move `avr-hal` to the new `avr-device` after #157 ... (For reference, I had already started on this in https://github.com/Rahix/avr-hal/pull/545 a while ago)

I'll leave this PR open until that work has concluded, I don't want to merge it without evidence that it produces the expected results.  Apart from that I think it's ready &mdash; thanks @LuigiPiucco for again picking up the work!

Rahix (approved):
Yay, managed to test this and it seems to work as expected! :tada: 

For reference, I used the following code:

```rust
#![no_std]
#![no_main]

use arduino_hal::prelude::*;
use panic_halt as _;

#[inline(never)]
fn stack_grower(level: usize, serial: &mut impl ufmt::uWrite) -> usize {
    let stack_size = avr_device::asm::get_stack_size();

    let next_size = if level < 16 {
        stack_grower(level + 1, serial)
    } else {
        stack_size
    };

    let _ = ufmt::uwriteln!(
        serial,
        "Stack At Level {}: {} (diff {})\r",
        level,
        stack_size,
        next_size - stack_size
    );

    stack_size
}

#[arduino_hal::entry]
fn main() -> ! {
    let dp = arduino_hal::Peripherals::take().unwrap();
    let pins = arduino_hal::pins!(dp);
    let mut serial = arduino_hal::default_serial!(dp, pins, 57600);

    ufmt::uwriteln!(&mut serial, "Hello from Arduino!\r").unwrap_infallible();

    stack_grower(1, &mut serial);

    loop {}
}
```

which produces the following output

```text
Hello from Arduino!
Stack At Level 16: 133 (diff 0)
Stack At Level 15: 125 (diff 8)
Stack At Level 14: 117 (diff 8)
Stack At Level 13: 109 (diff 8)
Stack At Level 12: 101 (diff 8)
Stack At Level 11: 93 (diff 8)
Stack At Level 10: 85 (diff 8)
Stack At Level 9: 77 (diff 8)
Stack At Level 8: 69 (diff 8)
Stack At Level 7: 61 (diff 8)
Stack At Level 6: 53 (diff 8)
Stack At Level 5: 45 (diff 8)
Stack At Level 4: 37 (diff 8)
Stack At Level 3: 29 (diff 8)
Stack At Level 2: 21 (diff 8)
Stack At Level 1: 13 (diff 8)
```

PR #198, opened 2025-04-30
build: Go back to the "legacy" identifier theme of svd2rust

### LINKED ISSUE #197: Switch back to svd2rust legacy identifier theme
After starting work on porting `avr-hal` to the new version of `avr-device`, I started hitting the effects of the changed identifiers in the generated API.  In <https://github.com/Rahix/avr-hal/pull/545>, commit <https://github.com/Rahix/avr-hal/pull/545/commits/ef69e1b1ddc6aafcbfda61372d9946a019d4cedc>, I at least got to the point of getting the Arduino Uno examples to compile again.

I have to say, I absolutely do not like the code churn we introduce with this.  While it is probably manageable for `avr-hal` with team effort, that's not all:

- All downstream users will have to work through the same churn of (sometimes non-trivial) identifier changes.
- All current PRs in `avr-hal` will need to be rebased by their authors to the new style identifiers.
- Legacy examples found on the web all become outdated.
- Rust compiler error messages on fixing the names are quite hit-or-miss.  The update strategy is not clear if you don't understand what has happened.

I took a look around the ecosystem and others have also decided to stick to what `svd2rust` calls the "legacy" identifier theme.  I am much in favor of doing the same.  There is no significant advantage to doing this change now, so let's not waste huge amounts of effor on it.  We already break enough other things with this update as it stands...

Cc: @LuigiPiucco @tones111

Rahix:
Resolves #197.

Cc: @LuigiPiucco @tones111

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -269,7 +269,7 @@ impl CodeGenerator {
         svd2rust_config.skip_crate_attributes = true;
         svd2rust_config.reexport_core_peripherals = false;
         svd2rust_config.reexport_interrupt = false;
-        svd2rust_config.ident_formats = IdentFormats::default_theme();
+        svd2rust_config.ident_formats = IdentFormats::legacy_theme();
 
         let generated_stream =
             svd2rust::generate::device::render(&svd, &svd2rust_config, &mut String::new())
--- a/examples/atmega328p/src/main.rs
+++ b/examples/atmega328p/src/main.rs
@@ -26,9 +26,9 @@ fn panic(_info: &core::panic::PanicInfo) -> ! {
 
     loop {
         avr_device::asm::delay_cycles(1_000_000);
-        dp.portd.portd().write(|w| w.pd3().set_bit());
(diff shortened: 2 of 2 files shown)


tones111:
Seems like a reasonable strategy as long as svd2rust continues to support the legacy naming convention.

The only counter argument I can think of is with AVR support maturing the number of dependent lines of code is likely to increase which will incur a larger switching cost the longer we wait.

I completely agree that it's a good idea to split the breakage so downstream users can work through each update independently.

tones111 (approved):

LuigiPiucco (approved):
I agree as well, since I used the new theme just because I didn't put much thought into it. One thing we should think about though, is that in case svd2rust decides to remove the legacy theme (or just preemptively), we should have a gradual migration plan. Maybe generate both themes (conditioned to a cargo feature), the new one under a different module path, and gradually switch over to it. Then some time after we swap them, so that the new theme becomes the default.

PR #207, opened 2025-06-02
Add AVR128DB28 device

tomtor:
Adds AVR128DB28

Note that the README currently has an "AtMega USB" column with two AVR64DU devices, which could be moved (or duplicated) to an AVR (Modern) column I added which now only contains AVR128DB28.

I checked the documentation and genererated code,  and I did not see the need for any yaml patch files (the AVR64DU devices also have none), but I might be missing something.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,6 +86,7 @@ attiny1614 = ["device-selected"]
 attiny1626 = ["device-selected"]
 avr64du32 = ["device-selected"]
 avr64du28 = ["device-selected"]
+avr128db28 = ["device-selected"]
 rt = ["avr-device-macros"]
 
 critical-section = ["critical-section/restore-state-u8"]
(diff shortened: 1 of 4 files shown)


Rahix:
> Note that the README currently has an "AtMega USB" column with two AVR64DU devices, which could be moved (or duplicated) to an AVR (Modern) column I added which now only contains AVR128DB28.

Yeah, the README table is a mess and it's probably best if I just get rid of it entirely soon...  If you want to, you can just drop you changes from the README entirely so there will be less merge conflicts.

> I checked the documentation and genererated code, and I did not see the need for any yaml patch files (the AVR64DU devices also have none), but I might be missing something.

We can always add more patches later if you don't see any immediate need for any.

Rahix (approved):
Thanks a lot!

PR #224, opened 2025-10-21
Add ATmega169(pa) support (v0.7.0 branch)

kolbma:
Not sure how to handle adding new controllers while avr-hal and avr-device diverged.  
I need a working/little bit stable API environment for this.  
So this PR is against the v0.7.0 codebase.  

Code for avr-hal/at-mega is coming, but need some more days for an example code using the LCD stuff.

### DIFF
--- a/patch/atmega169pa.yaml
+++ b/patch/atmega169pa.yaml
@@ -11,10 +11,15 @@ EXINT:
     _delete:
       - PCIE
     _add:
-      PCIE:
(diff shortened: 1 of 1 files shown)


Rahix:
avr-hal will soon be upgraded to the upcoming API of avr-device.  https://github.com/Rahix/avr-hal/pull/656 is the work-in-progress for this.

PR #228, opened 2025-11-14
Add nop functions to burn more than one cycle

mbuesch:
This PR adds a couple of more nop-like functions that burn more than one cycle.

The advantage of using these functions instead of using `delay_cycles` is that it's much more precise for a low amount of cycles.

The use case is that it's sometimes needed to sleep an extremely small amount of time. For example when bit banging fast busses or similar.
In such situations it's often not really suitable to delay a cycle more than necessary. For example if these situations happen many times in a row and therefore errors accumulate.

The same behavior could be achieved by calling the original nop() more than once in a row.
But the disadvantage is that this wastes more program memory space than necessary.

It's up to the user to protect this with IRQ-disable to avoid waiting more than necessary.

The nop functions are provided up to 5. From then on it starts to make sense to use `delay_cycles` instead.
Also these nop functions could be chained.

This PR also adds `options(preserves_flags)` to all asm blocks in this file. I think this makes sense as it doesn't change the memory or ordering behavior and avoids possibly unnecessary recomputations of flags.

### DIFF
--- a/src/asm.rs
+++ b/src/asm.rs
@@ -3,12 +3,116 @@
 #[cfg(target_arch = "avr")]
 use core::arch::asm;
 
-/// No Operation
+/// No Operation - Burn one CPU cycle
+///
+/// This emits exactly one NOP instruction and therefore burns at least one CPU cycle at runtime.
+///
+/// This function is an optimization fence.
+/// That means memory accesses will not be re-ordered by the compiler across this function call.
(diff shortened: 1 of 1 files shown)


Rahix:
> Do you plan to do a new release?

Yeah, really should get it done soon.  If I don't, feel free to annoy me more :)

PR #234, opened 2025-12-11
feat: add AVR128DB48 support

siku2:
This module is used as part of the [EV35L43A](https://www.microchip.com/en-us/development-tool/EV35L43A) eval kit.


Rahix (approved):
Thanks! :)

PR #250, opened 2026-03-26
Completed chip definitions for the mega0 series 

zacharytomlinson:
Completed chip definitions for the mega0 series focused on (3208/3209, 4808/4809)


zacharytomlinson:
@Rahix This was  the 'fix' for the issue regarding  confusion on 4809 implementation

PR #251, opened 2026-04-02
build: allow symlinks in required path validation (rules_rust compatibility)

stefanbucur:
Bazel's [rules_rust](https://github.com/bazelbuild/rules_rust) can fetch Cargo packages as dependencies, but in order to provide hermetic and reproducible builds, it creates symlinks for source files in sandboxed builds. The previous check in build.rs rejected these, causing builds to fail when using this crate with Bazel.

This change removes the symlink restriction so that symlinked paths are treated the same as regular files/dirs, enabling avr-device to work as a dependency in Bazel Rust repositories.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -154,11 +154,6 @@ impl InputFinder {
                 "required path `{}` is missing",
                 path.display()
             ));
-        } else if path.is_symlink() {
-            return Err(anyhow::anyhow!(
-                "required path `{}` being a symlink is invalid",
-                path.display()
-            ));
         }
 
         if path.is_dir() {


Rahix (commented):
Hi, thanks for the contribution! Making this crate work with bazel sounds very good.

I'm not quite following how those symlinks from bazel show up here.  We only call `track_path()` on directories inside the sources:

https://github.com/Rahix/avr-device/blob/5771519a3d244bf59046a6e3bda8ac4521818b73/build.rs#L73-L77

Does bazel replace these subdirectories with symlinks somehow?  What paths is it actually failing on?

I want to make sure we are not missing any odd side-effect of symbolic links here that could lead to us not properly tracking all necessary files.

Cc: @LuigiPiucco

LuigiPiucco:
I can't fully recall why symlinks were prohibited, it may have been just for completeness' sake. One issue I can see with allowing them is that, because directory traversal is recursive, it could lead to exploring many unrelated directories if the symlink is a directory and points to somewhere outside the project root (this could even be a circular link, in which case it would cause infinite looping). I'm not sure if that is a blocker, though.

However, the commit about `BUILD.bazel` seems unrelated (it's present in your other PR as well), and I'd expect a `BUILD.bazel` to be unneeded. If you're embedding avr-device into another Rust application that's built with Bazel, does it need to add a build file into the checked-out crate sources? Shouldn't that information be in the application's `BUILD.bazel`? Bazel supports cargo crates, right?

Maybe I'm misinterpreting the usage?

stefanbucur:
Thanks a lot for the quick responses!

> Does bazel replace these subdirectories with symlinks somehow? What paths is it actually failing on?

Yes, and the mechanism is as follows: Bazel decomposes the build graph into atomic build actions, where each action takes a set of files as inputs and produces output files. To ensure determinism and reproducibility for each action, Bazel creates a very lightweight ephemeral sandbox that consists of a working directory with symlinks to all the actual files the action needs.

For Rust crates, Bazel would download and extract the needed crates at the start of the build (as regular files), but will create symlinks to the crate files for each sandboxed action. My understanding is that for crates with custom build.rs scripts, the build action consists of running build.rs with a file layout that mimics the structure of the crate. So build.rs will see a similar structure as with Cargo-based unsandboxed builds, but all files are symlinks instead.

> However, the commit about BUILD.bazel seems unrelated (it's present in your other PR as well), and I'd expect a BUILD.bazel to be unneeded.

Thanks for the catch - I removed this part from the PR. This was a leftover from my fork of the repository. When Bazel builds crates, it first parses the Cargo.toml file and generates an equivalent BUILD.bazel file inside the directory where the crate archive is extracted. Since this directory is usually hidden from Bazel users, there should be no need to configure BUILD.bazel file ignores.

> If you're embedding avr-device into another Rust application that's built with Bazel, does it need to add a build file into the checked-out crate sources? Shouldn't that information be in the application's BUILD.bazel? Bazel supports cargo crates, right?

You're right - the user doesn't have to do anything. Bazel will automatically generate the BUILD.bazel for each crate based on the crate's Cargo.toml file.

Rahix:
I do have to say, this mechanism in bazel seems pretty cursed and fragile but I won't try arguing with it ^^

Happy to merge this, then!

Rahix (approved):

PR #252, opened 2026-04-02
feat: add ATtiny10 support

stefanbucur:
Add the ATtiny10 microcontroller (1 KB flash, 32 B SRAM, 4 GPIO pins). Includes the ATDF hardware description file, feature flag, module declaration, and all documentation/export updates following the same pattern as other supported devices. Peripherals: ADC, Analog Comparator, CPU, Port B, External Interrupts, 16-bit Timer/Counter (TC0), Watchdog Timer, Fuses. Package variants: SOT23-6, UDFN8.


Rahix:
Just out of curiosity, did you actually get a meaningful rust program to work with 32 bytes of SRAM?

stefanbucur:
> Just out of curiosity, did you actually get a meaningful rust program to work with 32 bytes of SRAM?

I actually did! That was my curiosity as well :) I was designing my first AVR board and wondered what's the smallest controller I can meaningfully place on the board. I was able to connect an ATtiny10 to a 3-to-8 demuxer and use 3 pins to strobe a 7-segment digit display using its timer interrupt. The rustc / LLVM size optimizations were decent, as well (I was able to get within the 10% code size overhead compared to the equivalent C on avr-gcc implementation).

I am just a beginner / hobbyist, but I can imagine this MCU being useful more generally in very low power settings where you need the minimum amount of silicon to implement some programmable logic, like maybe sampling from a sensor?

Rahix:
Very cool!

> I am just a beginner / hobbyist, but I can imagine this MCU being useful more generally in very low power settings where you need the minimum amount of silicon to implement some programmable logic, like maybe sampling from a sensor?

To be honest, I would always look for something ARM Cortex-M0 instead, just so much better supported and future-proof...

PR #259, opened 2026-07-07
[FEAT]: add avr128db32 device support

u-k-g:
## summary
- add the `avr128db32` device feature
- include the AVR128DB32 ATDF vendor file
- expose the generated `avr128db32` PAC module
- add AVR128DB32 to the crate docs and public reexports

## why
- enables PAC generation for AVR128DB32 targets
- supports boards using the 32-pin AVR128DB variant


Rahix (approved):
Thanks a lot!

PR #260, opened 2026-07-15
Update attiny412.atdf

fbradasc:
Update attiny412.atdf file from http://packs.download.atmel.com/Atmel.ATtiny_DFP.2.1.484.atpack

PR #261, opened 2026-07-17
Normalize tinyAVR 0/1-series ATDFs to one pack version

G33KatWork:
Unify all vendor ATDFs on Microchip ATtiny_DFP.3.4.278 pack. The old files were a mix of versions, which broke atdf2svd: some parts were older schema-0.3 files that encoded `TCA` as a `class="union"` register-group, an encoding atdf2svd cannot handle. It produced no `TCA0` in the PAC. The .278 pack encodes `TCA` as an ordinary register-group, so `TCA0` is generated correctly. attiny1614 was additionally a pre-release schema-0.3 file with wrong CCL pin data.

Also add patches for the 816, 1606 and 1614 so they match their siblings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ATtiny412`: Fixed `TCA0` missing and a bunch of incorrect register fields ([#260]).
 - Fixed compilation with the Bazel build system ([#251]).  There was a symlink
   check in our build script which interferes with Bazel's source organization.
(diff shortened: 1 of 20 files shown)


G33KatWork:
FYI, this is essentially the same as https://github.com/Rahix/avr-device/pull/260
I haven't checked every single datasheet and checked if the changes are correct like in the previous PR, however I am currently working on the atxtiny-hal again and adding support for more and more chips and while doing so, I found these discrepancies in the ATDFs between the newer xx17 chips and these. Then I found the previous PR and just updated all the ones that are the new tiny chips and fixed some patching as well.

PR #266, opened 2026-08-04
Unpin dev-dependencies.

Dirbaio:
Do not pin exact versions of dependencies. This creates version conflicts 
because it applies to the whole dependency tree, even if it's just build-deps. For example latest cortex-m 
requires a `syn` version higher than the pinned one, so Cargo bails out:

```
error: failed to select a version for `syn`.
    ... required by package `cortex-m-macros v0.7.8`
    ... which satisfies dependency `cortex-m-macros = "=0.7.8"` of package `cortex-m v0.7.8`
    ... which satisfies dependency `cortex-m = "^0.7.8"` of package `embassy-executor v0.10.0 (/home/dirbaio/embassy/embassy/embassy-executor)`
versions that meet the requirements `^2.0.117` are: 2.0.119, 2.0.118, 2.0.117

all possible versions conflict with previously selected packages

  previously selected package `syn v2.0.110`
    ... which satisfies dependency `syn = "=2.0.110"` of package `avr-device v0.8.1`
    ... which satisfies dependency `avr-device = "^0.8.1"` of package `embassy-executor v0.10.0 (/home/dirbaio/embassy/embassy/embassy-executor)`

failed to select a version for `syn` which could resolve this conflict
```

This means you can't make a crate that depends on both latest `avr-device` and `cortex-m`, which `embassy-executor` does.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -121,11 +121,11 @@ version = "=0.8.1"
 optional = true
 
 [build-dependencies]
-svd2rust = { version = "=0.37.1", default-features = false }
-svdtools = { version = "=0.5.0", default-features = false }
-atdf2svd = { version = "=0.5.1", default-features = false }
-prettyplease = "=0.2.37"
-svd-rs = { version = "=0.14.12", default-features = false }
-yaml-rust2 = { version = "=0.10.4", default-features = false }
-syn = { version = "=2.0.110", default-features = false, features = ["full", "parsing"] }
+svd2rust = { version = "0.37.1", default-features = false }
+svdtools = { version = "0.5.0", default-features = false }
+atdf2svd = { version = "0.5.1", default-features = false }
+prettyplease = "0.2.37"
+svd-rs = { version = "0.14.12", default-features = false }
+yaml-rust2 = { version = "0.10.4", default-features = false }
+syn = { version = "2.0.110", default-features = false, features = ["full", "parsing"] }
 anyhow = "1.0"


Rahix:
Hi!

Ugh, that's unexpected for me... I was under the impression that we can sidestep unification by pinning but it seems I misunderstood the rules once again :(

So the reason I would really like to have them pinned is that otherwise, I have to trust each of the dependencies to not change their behavior/generated output. I want the generated code of `avr-device` to stay deterministic for a given released version.

In the past, we have seen dependencies change behavior in patch releases, which led to the attempt to mitigate this.

Any thoughts or ideas? Just let loose and hope for the best? Build the codegen as a WASM blob with locked dependencies? I'm a bit at a loss here...

Dirbaio:
I don't think there's a good solution. Deps aren't *supposed* to do breaking changes in minor/patch releases so the "technically correct" solution is to not pin. The whole cargo house of cards is based on this assumption, that crates don't violate semver. But if you've already experienced semver violations...

Most PACs in the ecosystem just commit and ship the generated code which is disgusting but avoids this problem.

Rahix:
> Most PACs in the ecosystem just commit and ship the generated code which is disgusting but avoids this problem.

We did that, too, in the past, but switched to on-the-fly generation due to crate size...

Would you be happy if I unpin just `syn` as prepared in #267? At least `syn` I'd trust to do semver right. I know that's somewhat of a time-bomb of a solution, but at least it would fix the most likely problems?

Dirbaio:
yeah! that'd solve this particular problem. I'd suggest also unpinning `yaml-rust2`, keeping only the SVD stuff pinned maybe? `syn` and `yaml-rust2` are the more "general purpose" libs so it's more likely other crates will use them.

Rahix:
Alright, will do! I will cut a new release right after. It will be a semver break though since we already have breaking changes rolled up in `main`. Need a backport to `v0.8.x`? Or will you be fine to upgrade? (the breaking changes should not affect you I assume)

Dirbaio:
new release is fine, i'll just update it in embassy-executor. This means users will have to update `avr-device` but that's okay I guess.

Rahix:
Here you go: https://github.com/Rahix/avr-device/releases/tag/v0.9.0

